Issue #12, opened 2021-08-14
Requirements specify a single old version of seqeval


neubig:
Currently in `requirements.txt` there is a single, old version of seqeval specified:

    seqeval==0.0.12

In general it's probably better to avoid these "==" requirements, and only have ">" requirements, otherwise it can lead to environment conflicts.

PR #13, opened 2021-08-14
Updated requirements to remove seqeval version

neubig:
It seems that ExplainaBoard works with the most recent version of seqeval, so just removed the version specification.

Fixes https://github.com/neulab/ExplainaBoard/issues/12

Also makes a few small additions to `.gitignore`.


neubig:
Closing this in favor of a follow-up PR that will build upon the new naming of `interpret_eval` -> `explainaboard`

PR #43, opened 2021-09-25
Create template.json

pfliu-nlp:


neubig:
@pfliu-nlp : is this PR obsolete now?

PR #61, opened 2022-01-31
Reactivate continuous integration on github

neubig:
We had continuous integration setup on github, but it seems to have stopped working. Let's try again.


neubig:
This PR reactivates continuous integration. Tests are failing due to a missing dependency, but we can fix the tests in a different PR I think.

PR #64, opened 2022-01-31
fix: spacy setup and in-memory loaders

lyuyangh:
Migrated the PR from https://github.com/ExpressAI/ExplainaBoard/pull/35
1. Implemented `SpacyLoader`. 
    - It's intended for singleton use. The first time `spacy_loader.get_model("en_core_web_sm")` is called, it will load the model in memory. Subsequent calls will reuse the same model. 
    - It also prevents the need to load the spacy model in the global context which increases the latency for the web app significantly. With the PR, we only load the model when it's absolutely necessary so it's more predictable when it's going to cause latency (and when it's gonna fail). Also, just a heads-up, the web app containers we use in production only have 0.5G in memory (utilization is 25% when idle). If we plan to use big spacy models or other models, we need to consider allocating more memory.

2. Removed **requirements.txt** and moved spacy model download code to `install_requires`
    - I think it'll be easier if we put all the dependencies in one place (**setup.py**). The web app (and end-users who use SDK directly) install with `pip install explainaboard`. pip only installs dependencies specified in `install_requires`. For developers of the SDK, please install the package with `pip install .`.
    - Spacy models are a bit tricky. [Here](https://spacy.io/usage/models#download-pip) is the official documentation related to this. The models can be found in the releases of the spacy Github repo.

3. Implemented loaders for the `in_memory` type. I basically followed the code for the `local_filesystem` type. I didn't see any unit test for them so I couldn't test them fully. Please correct me if I made a mistake. 
    - Could you please implement `in_memory` types when you add a new `FileType`? Most of the time, `in_memory` should have pretty much the same implementation as the `local_filesystem` type. My original design was basically "when you load the file from the local file system, it becomes an in-memory type".

- One of the unit tests fail because (looks like) a service that the SDK uses is down?


pfliu-nlp:
>"Could you please implement in_memory types when you add a new FileType? "
No problem, good capture.

pfliu-nlp:
>"Removed requirements.txt and moved spacy model download code to install_requires"
It should be fine. 
We probably need to modify the [`ci`](https://github.com/neulab/ExplainaBoard/blob/main/.github/workflows/ci.yml) accordingly?

lyuyangh:
Thanks! I have updated CI to use `pip install .` . Also changed `python -m unittest discover tests` to `python -m unittest`. I think the default behavior does the correct thing for us. our tests directory is actually nested under **explainaboard/tests** so the CI was failing in the build stage. Now it seems to work except for that one test which seems to fail because **piaget.lti.cs.cmu.edu** is down.

pfliu-nlp:
>"Now it seems to work except for that one test which seems to fail because piaget.lti.cs.cmu.edu is down."
I will figure this out and try to solve it ASAP.

PR #89, opened 2022-02-11
feat: forbid duplicate version number in release ci

lyuyangh:
0.6.8 has been used previously so the release pipeline skipped it. Let's make it fail "loudly" whenever something like this happens. Otherwise, it's difficult to figure out what went wrong. Also, we'll never forget to update the version number again. What do you think?


pfliu-nlp:
Sounds reasonable, we can go with `skip_existing: false`, also to see if there are unexpected inconvenience.

neubig:
Just to confirm, are we releasing every time we incorporate a PR? If so, we might want to consider only releasing when we "tag" the PR as a release. For example, I did something like this in compare-mt here: https://github.com/neulab/compare-mt/blob/master/.github/workflows/publish-to-pypi.yml#L36

lyuyangh:
Yeah, I was just going to say something similar 😂. At this stage, I guess, most of the time, we want to release a new version for every PR because we want it to be reflected on the UI as soon as possible. We can maybe add the tagging mechanism after the public release? If there's something you want to hold off for some time (without releasing), you can leave the version file as it is. Then, the pipeline will fail. We can also modify the pipeline so it only runs if the version file changes (if having the pipeline fail in these scenarios is too annoying for everyone).

neubig:
I like the idea of the pipeline running only if the version file changes, that's a very simple and lightweight solution.

Issue #93, opened 2022-02-18
New Task: Support for MT


neubig:
We already support summarization, so support for MT should also be similar. We might also consider having a hierarchical level of support, like:
* sequence-to-sequence
  * summarization
  * MT
  * ...

Where there is a base set of features shared across all seq2seq tasks, and then specialized features for each sub-task.

In particular, it would be great if we could add some support for analyzing the datasets used in the class that I'm currently teaching, which would probably be helpful for the students in the class (although the deadline for the assignment is 2/25): http://phontron.com/class/multiling2022/assignment2.html

PR #95, opened 2022-02-19
Generalize conditional generation, add MT and examples

neubig:
Initial commit to add support for machine translation, fixes https://github.com/neulab/ExplainaBoard/issues/93

Basically this:
1. renames all of the "summarization" things into "conditional generation"
2. adds MT as another conditional generation task
3. adds examples and documentation for processing MT


explainaboard/processors/conditional_generation.py
```diff
@@ -61,4 +60,14 @@ def __init__(self, metadata: dict, system_output_data: Iterable[dict]) -> None:
...
+        self._builder = CondGenExplainaboardBuilder(self._system_output_info, system_output_data)
+
+
+@register_processor(TaskType.summarization)
+class SummarizationProcessor(ConditionalGenerationProcessor):
+    _task_type = TaskType.summarization
+
+
+@register_processor(TaskType.machine_translation)
+class MachineTranslationProcessor(ConditionalGenerationProcessor):
```
pfliu-nlp, line 72: All look good overall.
Notably, so far we assume that:
*  all [these features](https://github.com/neulab/ExplainaBoard/blob/17e8168c2688b24f2406f95b3d68b9e2e2d974b8/explainaboard/processors/conditional_generation.py#L14) are shared by all conditional text generation tasks. We can go with this first for immediate use, and refine them gradually (i.e., put more fundamental ones in the `ConditionalGenerationProcessor` class and task specifical one in `SummarizationProcessor` or `MachineTranslationProcessor`).
* We  regard `SummarizationProcessor`  as `text summarization` (since there will be other summarization tasks, such as video summarization.

explainaboard/tests/test_machine_translation.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def test_generate_system_analysis(self):
+        """TODO: should add harder tests"""
+
+        path_data = artifacts_path+ "test-mt.tsv"
+        loader = get_loader(TaskType.machine_translation, Source.local_filesystem, FileType.tsv, path_data)
+        data = loader.load()
+
+        metadata = {"task_name": TaskType.machine_translation.value,
+                    "dataset_name":"ted_multi",
```
pfliu-nlp, line 18: So far, we haven't token the languages into account, while when introducing the mt task, we probably need to consider it. (it seems that some evaluation metrics require the information of languages)

pfliu-nlp:
Left some minor comments and merged!

PR #96, opened 2022-02-20
Cleaned up some imports

neubig:
Some imports were done in a slightly unusual way, which caused PyCharm to give me a bunch of errors. This PR makes imports more standard.


lyuyangh:
Thank you for cleaning up the codebase! I'll wait for @pfliu-nlp 's comment on this. Do you think it'll be useful to enforce linting with something like flake8? We can enforce this automatically during the pre-commit stage and in the CI pipeline (we have something like this for the TS portion of the web app repo). btw, I think we have a "black" badge in the README too but this repository doesn't really use black😂. I can add a formatter (black) to the pipeline too if you guys agree on that.

Also, an issue I have struggled with when I work on this repository is that sometimes it can be difficult to find where the modules are referenced. I think it may be nice to avoid using wildcard imports (`from  xxx import *`) when it's not necessary. Another related issue is the use of `eval`. It is a really smart way of structuring our analysis code because it makes it flexible and easy to support new features and metrics. But the tradeoff is that I can't rely on IDEs to find references. And if we make a modification, it's difficult to know which modules it will affect. This works perfectly well now so we probably don't want to refactor it now. But I think it might be an issue we can consider in the future as it may become a barrier if we want people from the community to join and work on this project.

neubig:
Thanks @lyuyangh ! I agree with all of the above actually (formatting, import*, and eval). In particular, the use of `eval` is both a bit opaque, and seems like a little bit of a security risk.

explainaboard/builders/extractive_qa.py
```diff
@@ -1,22 +1,10 @@
...
-import sys
-sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
-from info import *
-import feature
 from typing import Iterable
-from info import SysOutputInfo
-from info import BucketPerformance
-from info import Performance
-from info import Table
+from explainaboard.info import SysOutputInfo, BucketPerformance, Performance, Table
```
pfliu-nlp, line 2: If also mentioning some details, I think this format would be better. lol
```python
from explainaboard.info import (
    SysOutputInfo, 
    BucketPerformance, 
    Performance, 
    Table,
)
```
It can reduce editing conflicts among users to some extent.

pfliu-nlp:
>"Do you think it'll be useful to enforce linting with something like flake8? We can enforce this automatically during the pre-commit stage and in the CI pipeline ... I can add a formatter (black) to the pipeline too if you guys agree on that."

Sounds good!

>"I think it may be nice to avoid using wildcard imports (from xxx import *) when it's not necessary."
Agree. But it maybe also depends on. For example, `utlis` -based scripts are usually used as "import *"

>"But the tradeoff is that I can't rely on IDEs to find references. And if we make a modification, it's difficult to know which modules it will affect. This works perfectly well now so we probably don't want to refactor it now. But I think it might be an issue we can consider in the future as it may become a barrier if we want people from the community to join and work on this project." 
yeah, it's kinda a problem left over by `history`. It's not in a higher priority in my list but could keep that in mind.

lyuyangh:
Sounds good! I resolved the merge conflict and I'm gonna merge this PR. I'll add flake8 and black when all the remaining PRs have been merged. It'll be nice to add those during a time when nobody is actively committing. otherwise, I'll introduce tons of merge conflicts. For Pengfei's comment about import syntax (multi-line import to avoid merge conflicts), I'll see if black or some other formatter enforces it by default. If not, we can probably add this custom rule.

PR #97, opened 2022-02-20
Feat/format & lint

lyuyangh:
A lot of files have been updated. It might be easier to review one commit at a time. Each commit accomplishes one task.
- Please reinstall the project
    ```
    pip install .[dev]
    pre-commit install
    ```
    - Please refer to README for instructions on how to set up flake8 and black in your dev environment. It will be very helpful if you configure your IDE to match our configurations. (I tested in VSCode and it's really easy. All you need to do is to select the correct linter and the correct formatter in settings. VSCode automatically locates the config files.)
- black
    - It should never break the code. I configured it to run in the pre-commit stage. It will format all files in place.
- flake8
    - Some flake8 rules are ignored because they conflict with black.
    - I modified the code to comply with flake8 if it's relatively easy to fix. For all other instances, I put `# noqa` to ignore the linter for now.
- Please let me know if you have found that a particular rule has made it difficult for you. We can add rules to ignore them.


neubig (approved):
Looks good to me although I'd also like @pfliu-nlp 's feedback. I think we should integrate this ASAP so we don't have merge conflicts with future commits. And thanks!

explainaboard/table_schema.py
```diff
@@ -37,160 +37,84 @@
...
-    },
-    {
-        "field_key": "true_label",
-        "sort": True,
-        "filter": True,
-        "label": "True Label"
-    },
+    {"field_key": "aspect", "sort": False, "filter": False, "label": "Aspect"},
+    {"field_key": "text", "sort": False, "filter": False, "label": "Text"},
+    {"field_key": "true_label", "sort": True, "filter": True, "label": "True Label"},
```
pfliu-nlp, line 113: It seems that there is a mysterious rule to judge when multiple-line format should be used. lol

neubig: Yeah, it's based on length (maybe 80 characters)? I think it's pretty standard.

pfliu-nlp:
Thanks Lyuyang and Graham. Looks good. I guess we also need a `black` and `flake8` for paper writing.

neubig:
Haha, great idea :)

lyuyangh:
> Thanks Lyuyang and Graham. Looks good. I guess we also need a `black` and `flake8` for paper writing.

Haha, yeah that would definitely be useful 😂

PR #99, opened 2022-02-22
add multiple-choice qa

pfliu-nlp:
* Since I have no system output file on these tasks, I just synthetically generate one with certail json format.
I will share this with Emmy who will give me a system output as planned.


pfliu-nlp:
ok, fixed the above issues.
the sdk for `qa-multiple-choice` basically can run, for example, using `gpt.json`, we can get:

image

PR #104, opened 2022-02-26
Check validity of inputs for pairwise analysis

neubig:
This checks to make sure that inputs are the same length for pairwise analysis, and also does a little bit of restructuring to reduce the amount of code in the main file. It will also be convenient when doing analysis across >2 systems if we decide to support that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ htmlcov
 explainaboard/tests/app.py
 explainaboard/tests/system_analysis.json
 system_analysis.json
+report.json
 
 
 log.res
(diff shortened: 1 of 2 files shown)


pfliu-nlp:
Looks nice. Merged. 
Also, @OscarWang114 , although currently, the the implementation of frontend's pair-wise analysis is independent of skd's, Graham's revision of this part would be also useful for the frontend stuff: https://github.com/neulab/ExplainaBoard/blob/2231ff6a2010ee2697b1641a9bedb67b70a0c292/explainaboard/explainaboard_main.py#L91
Also, do you think the the multiple system analysis part should be completely handed into sdk? If yes, what would be a good format if the report file with multiple system analysis?

neubig:
Yeah, the formatting for multiple system analysis is not straightforward, I think we'd have to think about this carefully.

Issue #105, opened 2022-02-27
Training set dependent and/or pre-calculated features are reported even when they don't exist


neubig:
Running the example from the doc on named entity recognition

    explainaboard --task named-entity-recognition --system_outputs ./data/system_outputs/conll2003/conll2003.elmo

It seems to be working well, other than the fact that entity-related feature buckets don't seem to be reported. 

    the information of #eCon#
    Bucket_interval	F1	Entity-Number
    [0.0,]	0.9222	5657
    
    the information of #eFre#
    Bucket_interval	F1	Entity-Number
    [0.0,]	0.9222	5657

This is apparently because training set dependent features are reported even when no training dataset is specified. This is confusing and probably should be fixed.

PR #107, opened 2022-02-28
Avoids printing missing features for the NER task

neubig:
Fixes https://github.com/neulab/ExplainaBoard/issues/105 and also does a little bit of code deduplication, refactoring, and typo correction.

I'd like @pfliu-nlp 's review, but also flagging @lyuyangh to quickly make sure this won't break the downstream interface (this means that some features are not output for some datasets).


lyuyangh:
Thanks! @OscarWang114 might be more familiar with this part. Could you take a look at the code when you get a chance? I think this should work but I'm not 100% sure. I can test these changes with the web app tonight.

neubig:
Sounds good! I reassigned it to @OscarWang114

OscarWang114:
@neubig @lyuyangh Thanks! The frontend uses the keys (the strings starting with `attr_` in the image) under the `results.fine_grained` field in the SDK outputted analysis dictionary to determine the number of graphs that should be plotted. So the only requirement is to ensure the un-outputted features are not included in `results.finegrained`. 

Screen Shot 2022-02-28 at 7 06 59 PM

I'm not entirely familiar with how the SDK outputs the analysis dictionary and might need some help with this. I can test out the code too, but probably in tomorrow afternoon due to a morning midterm. I hope the information above is clear and sufficient for @pfliu-nlp to review. Let me know if there are other questions!

neubig:
Yeah, it should be fine then! This PR basically avoids outputting the missing features so the frontend should just not print out the graphs based on this.

lyuyangh:
I tested this PR with the web app and it works for conll2003. But I don't think SDK should throw an error whenever it cannot find pre-computed features for a given dataset. The web app has no knowledge of whether the SDK has special features for a certain dataset so it sends the dataset name to SDK as long as the user chooses one. I think the SDK or DataLab SDK can maybe keep a list of dataset names that should have a pre-computed feature file. And it only throws an error if it cannot find the file for a dataset that's in the list? What do you think?

pfliu-nlp:
>"But I don't think SDK should throw an error whenever it cannot find pre-computed features for a given dataset."

Yeah, I kinda agree.

>"I think the SDK or DataLab SDK can maybe keep a list of dataset names that we support training set dependent features. And it only throws an error if it cannot find the file for a dataset that's in the list? What do you think?"

Good suggestion.  I think we probably can:
* try .. except the error from sdk side
* keep a list of datasets that support training set dependent features from sdk side
* what frontend needs to care about is to prompt users whether the dataset they used could be supported by training set dependent features

pfliu-nlp:
merged this branch and added `try except` for `FileNotFoundError`.

PR #112, opened 2022-03-09
Refactored and exposed pairwise performance gap

OscarWang114:
1. Refactored  `get_performance_gap` by moving it to a new folder called `analyzers` for accomodating future analysis functions. Renaming the function to `get_pairwise_performance_gap` for clarity.
2. Exposed `get_pairwise_performance_gap` so the web backend can use it.

Please let me know if I accidentally break anything, thanks!

Edit: Should I bump the version to 0.8.2?


pfliu-nlp:
Looks good!
It would be nice if we also calculate more useful statistics in `analyzer.py`, paving the way for frontend visualization or case analysis part.

PR #116, opened 2022-03-10
Start deduplicating code in builders classes

neubig:
There is a lot of copy-pasted code across the builders classes for different tasks, which makes the code verbose and a bit hard to maintain. This is a start of some deduplication by creating a parent class and moving shared functions there.

I'll continue with this in a future PR, but I think this is ready for review now!


pfliu-nlp:
Looks good, I will also put some time on deduplicating the current code.

PR #121, opened 2022-03-11
Fix/ci bug

lyuyangh:
Unfortunately, CI hasn't been working at all since I added flake8 and black🤦‍♂️. Now, all three checks are working (unittest, formatter and linter). I fixed all the formatting issues because that can be done automatically. @pfliu-nlp Do you mind fixing the unittests and the linter errors when you get a chance? Thanks!


neubig:
Tests seem to be running now, which was the intended effect of this commit. I'll go ahead and merge it.

Issue #125, opened 2022-03-11
Code deduplication -> code style


neubig:
Currently our flake8 tests for code style are failing on continuous integration, because the repository hasn't really been following coding conventions up till this point. I think it'd probably be a good idea to move in the direction of being compliant.

However, before we do that I think we should probably try to deduplicate the code so we don't end up fixing the formatting of the same code multiple times (which would be a bit of a waste). I'll try to take a lead on this but welcome any help!

PR #131, opened 2022-03-12
Finish "easy" deduplication of builders

neubig:
This abstracts the last of the "easy" functions in builders. Some builders (e.g. text classification and ABSA) now only define feature functions, which is nice.

FYI: my next steps are:

1. Doing some of the more non-trivial deduplication (where there are branching cases depending on the evaluation metric or single-metric vs. multi-metric, etc.)
2. I think now that the builders for each individual class are much simpler, the "processors" and "builders" can be merged together, so there's only a single file that users need to look at to understand how to implement a new task.
3. Code style: https://github.com/neulab/ExplainaBoard/issues/125


pfliu-nlp:
>"I think now that the builders for each individual class are much simpler, the "processors" and "builders" can be merged together, so there's only a single file that users need to look at to understand how to implement a new task."

Sounds good to me!

pfliu-nlp:
It seems that these lines have been removed: https://github.com/neulab/ExplainaBoard/blob/650319fcc22d31591c9cd15ee055173637b14434/explainaboard/builders/kg_link_tail_prediction.py#L190

Not sure if this your intention or something unexpected.

PR #133, opened 2022-03-14
Reduce stateful builders

neubig:
This is a large refactoring commit that attempts to reduce the "statefulness" of the builders. In other words, it refactors all of the functions to take clear inputs and return clear outputs, as opposed to adding information into the class member variables.

The main advantage of this is that it makes it easier for developers to understand the expected behavior of functions and the effects of running each function.

This is still WIP, but I've finished most of the design and the text classification test runs. I'd like to run it by @pfliu-nlp first before continuing implementing it for all builders. What do you think?

The most important file to look at is `explainaboard/builders/__init__.py` that contains the new interface.


pfliu-nlp (commented):
finish reviewing. Looks good!

Issue #135, opened 2022-03-15
Setting confidence score to None/null instead of zero if unavailable


OscarWang114:
Currently, it seems that the confidence scores (low and up) are set to zero if they are unavailable in the analysis. It would be better if we can set it to `None` in python and output `null` in JSON. It is more meaningful semantically and helps solve [this web interface bug](https://github.com/neulab/explainaboard_web/issues/111) I am fixing.

The bug is caused by the logic that confidence score points are only plotted when they are not zero (because my assumption is zero means they are not available). I missed the corner case where zero confidence scores can also be caused by a bucket with a score of zero. This could be fixed easily if SDK outputs `null` instead of zero.

neubig:
I can do this.

OscarWang114:
@neubig Thanks for the reply and taking on the issue! Looking forward to the PR. If at any moment you need to prioritize other matters, I'm happy to work on it too. 😀

PR #137, opened 2022-03-15
Set default of confidence scores to none

neubig:
Fixes https://github.com/neulab/ExplainaBoard/issues/135

Also, this renames `confidence_score_up` to `confidence_score_high` so it's congruent with `confidence_score_low`.

### DIFF
--- a/explainaboard/builders/conditional_generation.py
+++ b/explainaboard/builders/conditional_generation.py
@@ -239,13 +239,13 @@ class CondGenExplainaboardBuilder(ExplainaboardBuilder):
         for metric_name in sys_info.metric_names:
 
             overall_value = self.score_dic["corpus_level"]["corpus_" + metric_name]
-            confidence_score_low = 0.0
-            confidence_score_up = 0.0
+            confidence_score_low = None
+            confidence_score_high = None
             overall_performance = Performance(
                 metric_name=metric_name,
-                value=float(format(overall_value, '.4g')),
-                confidence_score_low=float(format(confidence_score_low, '.4g')),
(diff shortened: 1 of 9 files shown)


explainaboard/metric.py
```diff
@@ -51,19 +50,18 @@ def mean_confidence_interval(data, confidence=0.95):
...
 
         if not self._is_print_confidence_interval:
-            self._results["confidence_score_low"] = 0
-            self._results["confidence_score_up"] = 0
+            self._results["confidence_score_low"] = None
+            self._results["confidence_score_high"] = None
         else:
             (
                 confidence_interval_low,
                 confidence_interval_up,
```
OscarWang114, line 61: Let's rename `confidence_interval_up` as well.

explainaboard/builders/conditional_generation.py
```diff
@@ -239,13 +239,13 @@ def get_overall_performance(
         for metric_name in sys_info.metric_names:
 
             overall_value = self.score_dic["corpus_level"]["corpus_" + metric_name]
-            confidence_score_low = 0.0
-            confidence_score_up = 0.0
+            confidence_score_low = None
```
OscarWang114, line 242: Perhaps we can remove this variable and do (same for high):
```
Performance = (
...
confidence_score_low=None
...
)
```
This is more consistent with your code in other places. 😃

OscarWang114 (changes requested):
@neubig Thanks for the PR! I requested two small changes.

neubig:
Thanks! I'll fix the second one in #138 because it'll cause merge conflicts if I fix it here.

PR #138, opened 2022-03-16
Remove builders

neubig:
This PR removes builders from ExplainaBoard, consolidating everything into processors.

This forks off of https://github.com/neulab/ExplainaBoard/pull/137 so it can be merged after that.


neubig:
Formatting is failing but I'm not sure why. I think it might be a bug with Github CI, so maybe it can be fixed in a future commit.

neubig:
When I run python -m black . in the terminal, these files got fixed. This can also be fixed automatically by the pre-commit hook when you make a commit locally. Did you run pre-commit install to install the pre-commit hooks?

Thanks a lot! That fixed the issue, and it's nice to have the pre-commit hooks installed.

@pfliu-nlp : this is ready for a (light) review.

pfliu-nlp (approved):
done!

PR #140, opened 2022-03-17
feat: make ner faster

lyuyangh:
1. disable irrelevant spacy components when running ner
    - NER takes up 84% of the time when running text classification (got the number using `tests/artifacts/test-classification`). This change will reduce the overall latency by a little less than 50%. 
    - aspect-based sentiment classification also uses NER so it's also affected.
    - spaCy supports parallel processing out of the box so this can be improved further if we run NER for all samples at the same time. I didn't make this change because it makes the assumption that all the samples can be loaded into the memory. Do we want to leave room to handle datasets that are too large to be loaded into the memory at once? If not, it'll be easy to make some small changes to the processor so it can take advantage of the multiprocessing ability of spaCy.
    - before:
<br/>Screen Shot 2022-03-16 at 9 13 02 PM
    - after:
<br/>Screen Shot 2022-03-16 at 9 28 27 PM
    - to reproduce: `python -m cProfile -o text-classification-disable-irrelevant-spacy-components.pstats explainaboard/explainaboard_main.py --task text-classification --system_outputs /Users/lyuyangh/Developer/ExplainaBoard/explainaboard/tests/artifacts/test-classification.tsv --metrics F1score`

2. make eaas requests asynchronous for conditional generation tasks
    - `get_oracle` and the eaas request takes up 90% of the time. This change makes the eaas request execute on another thread so the overall latency is reduced by a little less than 50%.
    - The asyc logic is implemented in `AsyncEaaSClient`. If you think this is useful, I think we can move it to the eaas client itself. The current implementation uses threading which means there's a limit to the number of parallel requests it can make. I think a better solution would be to use coroutine which will scale a lot better. But it requires more changes to the SDK and the web app codebase because I need to turn a bunch of functions to be asyc. 
    - I didn't apply the same improvement to extractive QA because with the current design, parallelization wouldn't help.
    - before:
<br/>Screen Shot 2022-03-17 at 12 50 24 AM
   - after:
   <br/>Screen Shot 2022-03-17 at 12 56 51 AM
   - to reproduce: `python -m cProfile -o profile_results/mt-parallel.pstats explainaboard/explainaboard_main.py --task machine-translation --system_outputs cnndm_mini.bart --metrics bleu`
3. I also profiled this [larger file (cnndm_base)](https://drive.google.com/drive/folders/1TPE2M3R38_ugCAMJ05lEalqErkLhIeBr).
    - 75% of the time is spent on tokenization. Out of which 90% is on porter and the rest is mostly regex matching. I think this is kinda interesting because I didn't know stemming would take up all the time. Could we cache the stemming results so we can do a lookup for seen words? I guess we can also use multiprocessing here?
<br/>Screen Shot 2022-03-17 at 1 42 16 AM
   - reproduce: `python -m cProfile -o profile_results/mt-parallel-huge.pstats explainaboard/explainaboard_main.py --task machine-translation --system_outputs cnndm_base.tsv --metrics bleu`

### DIFF
--- a/explainaboard/processors/aspect_based_sentiment_classification.py
+++ b/explainaboard/processors/aspect_based_sentiment_classification.py
@@ -2,7 +2,7 @@ from explainaboard import feature
 from explainaboard.tasks import TaskType
 from explainaboard.processors.processor import Processor
 from explainaboard.processors.processor_registry import register_processor
-from explainaboard.utils.spacy_loader import spacy_loader
+from explainaboard.utils.spacy_loader import get_named_entities
 
 
 @register_processor(TaskType.aspect_based_sentiment_classification)
@@ -82,7 +82,6 @@ class AspectBasedSentimentClassificationProcessor(Processor):
 
     def __init__(self):
         super().__init__()
-        self._spacy_nlp = spacy_loader.get_model("en_core_web_sm")
 
     # --- Feature functions accessible by ExplainaboardBuilder._get_feature_func()
     def _get_sentence_length(self, existing_features: dict):
@@ -92,7 +91,7 @@ class AspectBasedSentimentClassificationProcessor(Processor):
         return len(existing_feature["text"])
 
     def _get_entity_number(self, existing_feature: dict):
-        return len(self._spacy_nlp(existing_feature["text"]).ents)
+        return len(get_named_entities(existing_feature["text"]))
 
     def _get_label(self, existing_feature: dict):
         return existing_feature["true_label"]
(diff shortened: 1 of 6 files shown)


neubig:
This is awesome thanks!

One comment is that it might actually be nice to have AsyncEaasClient as part of eaas instead of explainaboard in the future. @pfliu-nlp can make an executive decision, or I'm also planning on taking a closer look at the EaaS code soon as well and could make a suggestion then.

Issue #141, opened 2022-03-17
Separate dataset and system output


neubig:
Per discussion here: https://github.com/neulab/ExplainaBoard/issues/130

Currently dataset and system output are passed in as one file, but it would be better to pass them in separately, possibly with the dataset read directly from datalab.

PR #143, opened 2022-03-18
feat: extract user defined features code to the base loader

lyuyangh:
I was going through the loader code to prepare it for #141 when I found this "user-defined features" code that's specific to kg link tail prediction tasks. This PR extracts this code to the base loader so all task types support it. This also makes it easier to abstract loader logic to something more general like TSVLoader. Please let me know if this is useful!

- The configs dict for user-defined features are passed into the processors as part of the metadata but I couldn't find the place where it is used. Is it intended for some feature we want to support in the future or am I missing something?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ gen_samples.py
 log.res
 gen_samples.py
 explainaboard/tests/test_api.py
+profile_results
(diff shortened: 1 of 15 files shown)


pfliu-nlp (approved):
looks good , I also unittest the "user_defined_features" features on kg yesterday, it seems the result is not as our expectation but I need to check with Haris first.

PR #144, opened 2022-03-18
cleaning some folders

pfliu-nlp:


pfliu-nlp:
Thank to Yusuke's comments.

odashi:
@pfliu-nlp I guess some documents need to be fixed too. E.g., [supported_tasks.md](https://github.com/neulab/ExplainaBoard/blob/main/docs/supported_tasks.md) is using `data/`.

neubig:
I think it's OK to have a few datasets committed to the repo for demonstration purposes if they have an appropriate license that allows them to be distributed freely. But we should check on the datasets that we have right now to make sure that's the case. On the other hand, if it's really easy to download them from datalab we should just do that.

I agree that it'd be good to also think about what we need to do to fix the docs before merging this.

And after we delete the data/video, it'd also be nice to cleanse the repo history to reduce the size of the repo as well (particularly for the video). I'm not sure of the best way to do this in a clean way though.

odashi:
So it would be better to delete only the video for now? Only the way to remove the files permanently is to use `git filter-branch` I guess.

pfliu-nlp:
Hey, @odashi thanks for the comment!
>"I guess some documents need to be fixed too. E.g., [supported_tasks.md](https://github.com/neulab/ExplainaBoard/blob/main/docs/supported_tasks.md) is using data/."

Yes, but I actually haven't delete the folder `data`, just deleted `data/datasets`.

>"I agree that it'd be good to also think about what we need to do to fix the docs before merging this."
Just checked, it seems that we didn't use the files in the folder `data/datasets` I deleted.

>"But we should check on the datasets that we have right now to make sure that's the case. On the other hand, if it's really easy to download them from datalab we should just do that."

Also checked, It seems that we have already done this: there are some places where we assigned links to datasets, which either from huggingface datasets or datalab datasets.

neubig:
OK, I think we can go ahead with this then! We can clean the repo later.

PR #146, opened 2022-03-20
Start optimizing designs of conditional generation tasks [WIP]

pfliu-nlp:
Currently, the code for MT and summarization are entangled. This PR tries to reasonably decouple them


neubig:
LGTM. There's some dead code commented out instead of deleted, but we can fix that later given that #147 is dependent on this PR.

PR #148, opened 2022-03-21
added support for user-defined features for KG task

hwidjaja:
Previously, this was not working for the most recent refactored code.

Support was added for user-defined features for KG task ONLY; may need to discuss further how to generalize to the other tasks


explainaboard/processors/kg_link_tail_prediction.py
```diff
@@ -124,6 +124,35 @@ def process(self, metadata: dict, sys_output: List[dict]) -> SysOutputInfo:
         self._user_defined_feature_config = metadata.get(
             "user_defined_features_configs", None
         )
+
+        # add user-defined features into features list
```
hwidjaja, line 128: Ideally this part should be put inside the base processor, but the way that default features for each task are initialized, it's hard to do that

PR #153, opened 2022-03-21
Add tokenizers

neubig:
This adds a tokenizer class in `tokenizer.py` that can easily be generalized to other tokenizers later.

Consultation for @pfliu-nlp: there are currently hacky parts in the `get_statistics()` functions where I'm forced to hard-code a tokenizer in there because I'm not sure of the best way for the `Processor` class to pass an argument to `get_statistics()` (which is not a member function): https://github.com/neulab/ExplainaBoard/blob/add_tokenizer_class/explainaboard/processors/conditional_generation.py#L293

Do you know if it would be possible to move the `get_statistics` function to be a member function of `Processor`? That would make it possible for it to use member variables of Processor.

(I think this PR can be merged before we do that though, as that would be a bigger change)

### DIFF
--- a/explainaboard/processors/aspect_based_sentiment_classification.py
+++ b/explainaboard/processors/aspect_based_sentiment_classification.py
@@ -85,7 +85,7 @@ class AspectBasedSentimentClassificationProcessor(Processor):
 
     # --- Feature functions accessible by ExplainaboardBuilder._get_feature_func()
     def _get_sentence_length(self, existing_features: dict):
-        return len(existing_features["text"].split(" "))
+        return len(self._tokenizer(existing_features["text"]))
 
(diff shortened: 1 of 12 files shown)


pfliu-nlp:
Thanks, Graham.

>"Do you know if it would be possible to move the get_statistics function to be a member function of Processor? That would make it possible for it to use member variables of Processor."

I think it could be doable, I will explore a good method to achieve this goal.
Regarding how to introduce additional arguments for `get_statistics` function,   NER processor actually has done this: https://github.com/neulab/ExplainaBoard/blob/2ff18e2c1f8604c1cb230de49ed40ea9a489366a/explainaboard/processors/named_entity_recognition.py#L803

Issue #134, opened 2022-03-14
update the doc accordingly


pfliu-nlp:
dont' forget to update the [doc](https://github.com/neulab/ExplainaBoard/blob/main/docs/add_new_features.md) accordingly (w.r.t `bucketInfo`)

PR #157, opened 2022-03-22
Updated doc about new features and tasks

neubig:
Fixes https://github.com/neulab/ExplainaBoard/issues/134

Also removes `table_schema.py` which I believe is out-of-date now?


OscarWang114:
@neubig Thanks for the PR! `table_schema.py` is actually required by the web interface to display the system output table in a task-dependent way (e.g. marking specific columns as sortable). I have not implemented this feature so it hasn't been used. Sorry if that causes the confusion!

neubig:
Could we instead get that information by parsing `Processor._features`?

OscarWang114:
Ah, preventing duplication is a great point. The two pieces of information are not identical though. `Processor._features` is storing feature information for processing (e.g. is_bucket, bucket_info), while `table_schema.py` is storing a subset of fields from `Processor._features` with additional UI-related information for the web interface (e.g. sort: False).

We can move the information into `Processor` by creating a class member called `_table_schemas` for each processor. So `_features` is used for processing, and `_table_schemas` is used for UI. While there's no duplicate elimination, the two pieces of information will be organized in one place. I'm happy to do so you think it's good! The only concern is storing UI information isn't really the processor's responsibility.

If needed in the future, we can even merge `_features` and `_table_schemas`, to eliminate duplication further or extract them to a separate class named `Features` for every task.

neubig:
Alternatively, if this is only used by the web frontend, maybe it could be included in the web frontend code instead?

The reason why I thought it was obsolete is because it's not used by any other code in this repo, which may indicate that it should be placed somewhere else?

I don't care too strongly though, I'm also OK with undeleting it and keeping it here.

OscarWang114:
That makes sense, too. Let me open an issue and we can keep it for now?

neubig:
Thanks, I restored the files for now.

PR #162, opened 2022-03-22
Separate out conditional generation scoring

neubig:
As I started working on the conditional generation code to expand its functionality I realized it was saving `self.score_dict`, which made the processor stateful (and complicates the saving of statistics to the DB, as @OscarWang114 is working on). This PR splits off the calculation of scoring functions into a separate function.

### DIFF
--- a/explainaboard/processors/conditional_generation.py
+++ b/explainaboard/processors/conditional_generation.py
@@ -106,12 +106,14 @@ class ConditionalGenerationProcessor(Processor):
             existing_features, statistics, lambda x: x['source']
         )
 
-    def _complete_features(
-        self, sys_info: SysOutputInfo, sys_output: List[dict], external_stats=None
-    ) -> List[str]:
-        """
-        This function is used to calculate features used for bucketing, such as sentence_length
-        :return:
+    def _gen_scoring_stats(
+        self, sys_info: SysOutputInfo, sys_output: List[dict]
+    ) -> Any:
+        """Generate sufficient statistics for scoring.
+
+        :param sys_info: Information about the system outputs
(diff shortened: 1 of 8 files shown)


explainaboard/processors/conditional_generation.py
```diff
@@ -106,12 +106,14 @@ def _get_fre_rank(self, existing_features: dict, statistics: Any):
...
 
-    def _complete_features(
-        self, sys_info: SysOutputInfo, sys_output: List[dict], statistics=None
-    ) -> List[str]:
-        """
-        This function is used to calculate features used for bucketing, such as sentence_length
-        :return:
+    def _gen_scoring_stats(
+        self, sys_info: SysOutputInfo, sys_output: List[dict]
+    ) -> Any:
```
OscarWang114, line 111: It seems the function is returning a dictionary. We could be more explicit by using dict Instead of Any.

OscarWang114 (commented):
Thanks for the helpful PR! I added some comments.

explainaboard/processors/conditional_generation.py
neubig: I debated about this, but I think theoretically the scoring stats could be in a different format. For example, a numpy array would be a good option. Maybe we can leave it as "Any" for now.

neubig:
Thanks for the comments, @OscarWang114 ! I believe they've been resolved unless you feel strongly about the typing issue.

explainaboard/processors/conditional_generation.py
OscarWang114: Ah I see, then Any is fine!

PR #163, opened 2022-03-22
[WIP] Split content in process(), expose _bucketing_samples()

OscarWang114:
Update: Ready for review!

As the leaderboard in the web frontend needs an overall score for each system, I will also be saving the results from `get_overall_performance` in addition to sufficient statistics. 

Two functions are created:
1. `get_overall_statistics` for getting sufficient statistics and overall performance. It is also called by `process()` for reusability. For the web interface, the returned results are saved to the DB.
2. `get_fine_grained_statistics` exposes `_bucketing_samples` for the web interface to call on the fly, since we're not supposed to call underscore functions from outside.

Let me know if you have any concerns!

### DIFF
--- a/explainaboard/processors/processor.py
+++ b/explainaboard/processors/processor.py
@@ -384,7 +384,9 @@ class Processor:
         active_features = self._complete_features(
             sys_info, sys_output, external_stats=external_stats
         )
-        overall_results = self.get_overall_performance(sys_info, sys_output)
+        overall_results = self.get_overall_performance(
+            sys_info, sys_output, scoring_stats=scoring_stats
+        )
         return {
             "sys_info": sys_info,
             "scoring_stats": scoring_stats,
--- a/explainaboard/processors/processor.py
+++ b/explainaboard/processors/processor.py
@@ -1,5 +1,5 @@
 import json
-from typing import Callable, List, Tuple, Dict
+from typing import Callable, List, Tuple, Dict, Optional
 
 from datalabs import load_dataset
 from explainaboard.utils.async_eaas import AsyncEaaSClient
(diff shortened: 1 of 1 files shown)


explainaboard/processors/processor.py
```diff
@@ -356,6 +356,13 @@ def process(self, metadata: dict, sys_output: List[dict]) -> SysOutputInfo:
         active_features = self._complete_features(
             sys_info, sys_output, statistics=statistics
```
pfliu-nlp, line 364: it seems that we have multiple types of [`statistics`](https://github.com/neulab/ExplainaBoard/pull/162/commits/d863afd2e416a038a2cf4800cc9912f727c44c4d), do we want to make the name of this `statistics` more clear?

neubig: I think that's fixed in the other PR, which @OscarWang114 will still need to merge into this one.

OscarWang114: Thanks. Yes, I will merge them now. Also, this is still a draft so just let me know if the high-level direction sounds good, I will convert it to a formal one once the details are properly implemented.

pfliu-nlp: thanks! it looks good and I will be ready whenever needed to be reviewed.

OscarWang114: Thanks! It's ready now.

pfliu-nlp: Thanks will do it now.

explainaboard/processors/processor.py
```diff
@@ -35,11 +37,19 @@ def __init__(self) -> None:
...
-        of statistics.
+    def _get_statistics_resources(
+        self, dataset_split: Dataset
+    ) -> Optional[Mapping[str, Any]]:
+        """
+        From a DataLab dataset split, get resources necessary to calculate statistics
+        """
+        return None
+
+    def _gen_external_stats(
```
pfliu-nlp, line 48: there is a `statistics` inside this function,  would it be a good idea to rename them as `external_stats` in the future?

pfliu-nlp (commented):
overall it looks good

pfliu-nlp:
Overall it looks good, and I particularly care about the concept of `statistics`, I merged this PR first and we can make other relevant discussion here: https://github.com/neulab/ExplainaBoard/issues/164

OscarWang114:
Thanks @pfliu-nlp! I'll take a look at the discussion later today.

PR #165, opened 2022-03-23
generalize customized features & multiple-choice QA as case study

pfliu-nlp:
This PR tries to generalize the implementation of customized features to more tasks, specifically
- [x] add a  `_init_customized_features` member function into `processor` class
- [x] generate test set for multiple-choice QA task artificially and test it 
```
explainaboard --task qa-multiple-choice --system_outputs ./data/system_outputs/metaphor_qa/synthetic.json > report.json
```


pfliu-nlp:
Notably:

(1) when deploying in the web platform,  it seems that we should explicitly specify this: https://github.com/neulab/ExplainaBoard/blob/4be00f2a21fc4c668a03872a217933d4018a8d2f/explainaboard/tests/test_qa_multiple_choice.py#L54
Otherwise, user-defined features will be ignore.

(2) Regarding the `multi-label` feature, currently SDK will concatenate them, which probably will be generalized in the feature.
https://github.com/neulab/ExplainaBoard/blob/4be00f2a21fc4c668a03872a217933d4018a8d2f/data/system_outputs/metaphor_qa/synthetic.json#L25

image

explainaboard/processors/processor.py
```diff
@@ -413,6 +468,7 @@ def get_fine_grained_statistics(
         }
 
     def process(self, metadata: dict, sys_output: List[dict]):
+        # TODO(Pengfei): Rethink if this is a good way to manipulate `system_output`
```
neubig, line 471: FWIW, I agree. I wondered about this myself when implementing this in this way.

neubig (approved):

neubig:
Looks great. Just one small comment: there was a place where you used Google format docstrings:

    Args:
       XXX: ...

    Returns:
      ...

But I think that in general we use reST doctstrings, so this may be worth fixing.

     :param XXX: ...
     :returns: ...

Reference here:
https://stackoverflow.com/questions/3898572/what-are-the-most-common-python-docstring-formats

I don't think that's worth blocking the PR though so I went ahead and merged.

pfliu-nlp:
> Looks great. Just one small comment: there was a place where you used Google format docstrings:

Oh, I see.  I used it since it's auto-generated by pycharm. -_-!

PR #167, opened 2022-03-23
Refactor/loaders

lyuyangh:
1. Commit 1: refactored `Loader.__init__()`
  - made `data` a required argument
  - all loaders now call the `__init__` method of the base loader
2. Commit 2: implemented file-specific loaders to simplify the task-specific loaders 
  - implements `TSVFileLoader`, `JSONFileLoader`, `DatalabFileLoader` and `CoNLLFileLoader` which knows how to load a certain type of file given the fields
  - refactored all the existing loaders to use these file-specific loaders instead
  - `QAMultipleChoiceLoader` `KgLinkTailPredictionLoader` still uses custom `load()` methods because they support user-defined features. The way they load these extra features is different so I decided to leave them for now. It'll be easy to incorporate user-defined features to the file loaders (we just need to update the fields based on `self.user_defined_features_configs`)
  - hellaswag is removed in https://github.com/neulab/ExplainaBoard/commit/4b93b9542b714754eb91d718cd82b98ab706d11c
  - This refactor makes it easier to do #141 in the future. We just need to have two sets of file loaders for each task-specific loader. One is for the (input, reference_output) file and the other one is for the predictions file.

Please let me know what you think! Thanks!


explainaboard/loaders/file_loader.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def __post_init__(self):
+        for name in (self.src_name, self.target_name):
+            if not isinstance(name, str) and not isinstance(name, int):
+                raise ValueError("src_name and target_name must be str or int")
+        if self.dtype == FileLoaderDType.other and self.strip_before_parsing:
+            raise ValueError("dict type field does not support strip_before_parsing")
+
+
+class FileLoaderDType(str, Enum):
```
neubig, line 110: Do we need this? Could we just use the standard Python typing mechansim?

explainaboard/loaders/aspect_based_sentiment_classification.py
```diff
@@ -15,51 +20,31 @@ class AspectBasedSentimentClassificationLoader(Loader):
...
-                )
-        else:
-            raise NotImplementedError
-        return data
+    _default_file_type = FileType.tsv
+    field_names = ["aspect", "text", "true_label", "predicted_label"]
+    _default_file_loaders = {
+        FileType.tsv: TSVFileLoader(
+            [
+                FileLoaderField(0, field_names[0], FileLoaderDType.str, True),
```
neubig, line 28: The first three fields here are pretty obvious from the context, but the final True is not. Could you use `strip_before_parsing=True` instead to make this more obvious?

explainaboard/loaders/file_loader.py
```diff
@@ -0,0 +1,221 @@
...
+            parsed_data_points.append(parsed_data_point)
+        return parsed_data_points
+
+
+@dataclass
+class FileLoaderField:
+    src_name: Union[int, str]  # int for tsv column indices and string for dict keys
+    target_name: str
+    dtype: FileLoaderDType
+    strip_before_parsing: bool = False
```
neubig, line 99: It wasn't immediately obvious to me what this meant. A comment would be helpful.

Also, should we make True be default? It seems this is used far more often than False.

explainaboard/loaders/file_loader.py
```diff
@@ -0,0 +1,221 @@
...
+            raise ValueError("target_name must be unique")
+
+    @staticmethod
+    def parse_data(data: Any, field: FileLoaderField) -> Any:
+        if field.parser:
+            return field.parser(data)
+        if field.strip_before_parsing:
+            data = data.strip()
+        dtype = field.dtype
+        if dtype == FileLoaderDType.int:
```
neubig, line 42: `if` and `elif` finally `else` may be better? Can probably also include the `if field.parser` statement.

explainaboard/loaders/file_loader.py
```diff
@@ -0,0 +1,221 @@
...
+        if len(src_names) != len(set(src_names)):
+            raise ValueError("src_name must be unique")
+        if len(target_names) != len(set(target_names)):
+            raise ValueError("target_name must be unique")
+
+    @staticmethod
+    def parse_data(data: Any, field: FileLoaderField) -> Any:
+        if field.parser:
+            return field.parser(data)
+        if field.strip_before_parsing:
```
neubig, line 39: Shouldn't this before the previous if statement, esp given the naming?

explainaboard/loaders/file_loader.py
```diff
@@ -0,0 +1,221 @@
...
+                raise ValueError("field src_name for TSVFileLoader must be an int")
+            if field.dtype == FileLoaderDType.other:
+                raise ValueError("TSVFilerLoader doesn't support 'other' field type")
+
+    @classmethod
+    def load_raw(self, data: str, source: Source) -> Iterable:
+        if source == Source.in_memory:
+            file = StringIO(data)
+            return csv.reader(file, delimiter='\t')
+        if source == Source.local_filesystem:
```
neubig, line 136: elif, else

explainaboard/loaders/file_loader.py
```diff
@@ -0,0 +1,221 @@
...
+
+class CoNLLFilerLoader(FileLoader):
+    def __init__(self) -> None:
+        super().__init__(fields=[], use_idx_as_id=False, id_field_name=False)
+
+    @classmethod
+    def load_raw(self, data: str, source: Source) -> Iterable:
+        if source == Source.in_memory:
+            return data.splitlines()
+        if source == Source.local_filesystem:
```
neubig, line 153: elif, else.

neubig (changes requested):
This looks awesome! I left several comments.

explainaboard/loaders/file_loader.py
lyuyangh: This is a great suggestion! I replaced the enum type with the python types (int, float, str, and None). None means other. I originally added an enum type because I thought we would need to support some more advanced types but I think these simple types cover everything we need now. If a user wants to work with some more complicated data schema, they can write their own parser pretty easily.

explainaboard/loaders/file_loader.py
lyuyangh: I added a docstring for `FileLoaderField`. I made `strip_before_parsing=True` to be the default for `dtype==str` only. I didn't want to make this argument to be True by default because it only works if the src data is a string. If a user ignores this field and left it to be True, they may get an error message about `strip()` which may not be obvious what that means. But I think this should be relatively safe if the target data type is str.

explainaboard/loaders/file_loader.py
lyuyangh: `parser` is intended for custom parsers. If a custom parser is provided, everything else will have no effect (dtype type conversion and `strip_before_parsing`). I think this behavior may be more intuitive? because if `strip()` is called on something other than str, the user will get an error message and it may not be obvious what causes that error.

lyuyangh:
Thank you for all the comments! I resolved most of them except for the parser one. I think it would be more intuitive for the user if I disable everything else if a custom parser has been provided. Please let me know what you think!
I added a unit test for hellaswag but it's failing in the CI. I'm not sure what that error message means. I'll look into this tomorrow. Please let me know if any of you can reproduce this locally.

pfliu-nlp:
Thanks for all commits, will review them tomorrow!

neubig (approved):

neubig:
For @pfliu-nlp : do we need the special hellaswag class still? I think we could probably handle the hellaswag dataset under `MultipleChoiceQA`? If so, we could just delete all the things regarding hellaswag to fix the codebase. If that's the case we can just ignore this error and delete the hellaswag stuff instead.

If we do need to fix it, I was able to reproduce the error locally. I wonder if this issue may be of help?
> https://github.com/huggingface/transformers/issues/11021

explainaboard/loaders/file_loader.py
```diff
@@ -0,0 +1,221 @@
...
+        fields: List[FileLoaderField] = None,
+        use_idx_as_id: bool = True,
+        id_field_name: Optional[str] = None,
+    ) -> None:
+        super().__init__(fields, use_idx_as_id, id_field_name)
+        for field in self._fields:
+            if not isinstance(field.src_name, int):
+                raise ValueError("field src_name for TSVFileLoader must be an int")
+            if field.dtype == FileLoaderDType.other:
+                raise ValueError("TSVFilerLoader doesn't support 'other' field type")
```
pfliu-nlp: Is this (`TSVFiler`) a typo?

explainaboard/loaders/file_loader.py
```diff
@@ -0,0 +1,221 @@
...
+        if source == Source.local_filesystem:
+            content: List[str] = []
+            with open(data, "r", encoding="utf8") as fin:
+                for record in csv.reader(fin, delimiter='\t'):
+                    content.append(record)
+            return content
+        raise NotImplementedError
+
+
+class CoNLLFilerLoader(FileLoader):
```
pfliu-nlp, line 160: Similarly, is this a typo?

explainaboard/loaders/named_entity_recognition.py
```diff
@@ -1,4 +1,4 @@
-from typing import Dict, Iterable, List
+from explainaboard.loaders.file_loader import CoNLLFilerLoader
```
pfliu-nlp, line 1: mark

explainaboard/loaders/named_entity_recognition.py
```diff
@@ -17,51 +17,4 @@ class NERLoader(Loader):
...
-        data.append(
-            {
-                "id": str(guid),
-                "tokens": tokens,
-                "true_tags": ner_true_tags,
-                "pred_tags": ner_pred_tags,
-            }
-        )
-        return data
+    _default_file_loaders = {FileType.conll: CoNLLFilerLoader()}
```
pfliu-nlp, line 20: mark

pfliu-nlp:
> For @pfliu-nlp : do we need the special hellaswag class still? I think we could probably handle the hellaswag dataset under `MultipleChoiceQA`? If so, we could just delete all the things regarding hellaswag to fix the codebase. If that's the case we can just ignore this error and delete the hellaswag stuff instead.
> 
> If we do need to fix it, I was able to reproduce the error locally. I wonder if this issue may be of help?
> 
> > [huggingface/transformers#11021](https://github.com/huggingface/transformers/issues/11021)

I think removing the `hellaswag` would be a better choice for now.

lyuyangh:
@pfliu-nlp Thanks! And yeah no rush. I removed hellaswag. If someone wants to reference the code in the future, the commit hash is 4b93b9542b714754eb91d718cd82b98ab706d11c

explainaboard/loaders/file_loader.py
neubig: OK, sounds good.

pfliu-nlp:
Some high-level summary of undersstanding:
* `Loader.loader()`: task-aware loader
* `FileLoader.load_raw()`: modality-aware loader (e.g., in-memory, local file system)
* `FileLoader.load()`: format-aware loader (e.g., tsv, conll)

Relationship:
```python
Loader.loader():
     FileLoader.load():
          Fileloader.load_raw():
```
or  (for some tasks, such as [kg_link_tail_prediction](https://github.com/neulab/ExplainaBoard/blob/refactor/loaders/explainaboard/loaders/kg_link_tail_prediction.py)
```python
Loader.loader():
       Fileloader.load_raw():
```

I'm thinking about
* do we need this `if`: https://github.com/neulab/ExplainaBoard/blob/4b93b9542b714754eb91d718cd82b98ab706d11c/explainaboard/loaders/kg_link_tail_prediction.py#L30
* or should this `low_raw` be conducted before the above `if`: https://github.com/neulab/ExplainaBoard/blob/4b93b9542b714754eb91d718cd82b98ab706d11c/explainaboard/loaders/qa_multiple_choice.py#L33

pfliu-nlp:
I think I'm fine with merging this PR, we can go with it to see how generalized it is. Further modifications could be made if needed in the future.

pfliu-nlp (approved):

lyuyangh:
@pfliu-nlp Thanks! And yeah, `FileLoader` is another level of abstraction so it has made the code a little harder to understand. I went back and forth with the design a couple of times when I implemented this. I think your high-level summary is correct. Here're some additional comments:

1. All the methods in the file loaders are aware of the file format. `FileLoader.load_raw()` is responsible for "loading data from source and return an iterable of unparsed data points (the data type and the format of those data points depend on the file type)." (from the docstring). `load_raw()` is a class method so it does not have any knowledge of `self._fields`. It only knows which class it is in so it knows about the file type.  Whereas `FileLoader.load()` uses `self._fields` to not only load data into an iterable of dict. The keys and value types of the dictionary is defined by `self._fields`.
2. `FileLoader.load()` calls `FileLoader.load_raw()`. The reason why I separated the two methods are:
    1. Separating concerns. Most of the time, you don't need to override `FileLoader.load()` from the base `FileLoader`. If you want to support a new file format, all you need to do is to override `FileLoader.load_raw()`. `CoNLLFileLoader` is the only exception at the moment.
    2. Making it easier if some task loaders (derived class of `Loader`) need to define custom `Loader.load()` methods. In those situations, you can still use `FileLoader.load_raw()` to get the iterable. Otherwise, you will have to implement that too.
3. `Loader` can have multiple `FileLoader`s.
    - If we decide to separate the inputs and outputs like we discussed today during the meeting. Each `Loader` would have two attributes: `self._input_loaders` and `self._output_loaders`. And `Loader.load()` could do something like this:
    ```
    input: Iterable[dict] = self._input_loaders[input_file_type].load(input_path)
    ref_output: Iterable[dict] = self._output_loaders[ref_output_type].load(ref_output_path)
    predicted_output: Iterable[dict] = self._output_loaders[predicted_output_type].load(predicted_output_path)
    sys_output = merge(input, ref_output, predicted_output)
    ```
    - we can separate the sys_output file some other way but I think the general idea would be similar to this. If we don't have these `FileLoaders`, we may need to write a lot more code to achieve something similar.
4. Regarding your comment about `KgLinkTailPredictionLoader.load()`. I made a couple of mistakes there.
    - I should have used `self.file_loaders` instead of `self._default_file_loaders`. The default file loaders are the defaults for this task. If someone wants to load a custom dataset they have the option to provide custom file loaders when they initialize `Load`.
    - You are right. We don't need to check the file type here because it has been validated in `Loader.__init__()`. https://github.com/neulab/ExplainaBoard/blob/0d6e38c79c493c30eedbeca82a9cff641212a002/explainaboard/loaders/loader.py#L43-L46

pfliu-nlp:
Hi, @lyuyangh thanks a lot for such detailed replies.

>"I went back and forth with the design a couple of times when I implemented this. I think your high-level summary is correct. Here're some additional comments:"

Haha, yesyes, I can understand it, and designing the data loader is like writing an introduction of a research paper, which always need multiple revision.

>"If you want to support a new file format, all you need to do is to override FileLoader.load_raw(). CoNLLFileLoader is the only exception at the moment."

I'm thinking that could we move some shared parts (e.g., reading from `in_memory` or `local_filesystem`) into the `FileLoader.load_raw()` instead of raising ` NotImplementedError`
https://github.com/neulab/ExplainaBoard/blob/4b93b9542b714754eb91d718cd82b98ab706d11c/explainaboard/loaders/file_loader.py#L69

>"Loader can have multiple FileLoaders.."
This makes sense to me and also seems pretty closed to what Graham has suggested.

>"Regarding your comment about KgLinkTailPredictionLoader.load(). I made a couple of mistakes there."
Thanks!

lyuyangh:
> I'm thinking about if we could move some shared parts (e.g., reading from in_memory or local_filesystem) into the FileLoader.load_raw() instead of raising NotImplementedError

I don't think there are any shared parts because `load_raw()` needs to know the file type (different file types would require different code). Also, this design makes it easier for the developers to understand what needs to be done when implementing a derived class of `FileLoader`. Raising `NotImplementedError` in `FileLoader.load_raw()` basically makes it an abstract method so it forces the developers to override it in the derived classes.

pfliu-nlp:
>"because load_raw() needs to know the file type (different file types would require different code)"
if this is the case, I agree!
I have no other questions at the time being, let's merge it.

PR #168, opened 2022-03-23
this PR refactors `get_statistics`, making it as member function

pfliu-nlp:
This PR aims to make `get_statistics` as a member function (relatively complicated than I expected, I need to modify something of DataLab), together fix some bugs that haven't been identified by existing unittests.


pfliu-nlp:
This PR is ready for merging. In the future,
* more unit-tests for training set dependent features should be added (I guess there are still some hidden bugs)

neubig (approved):
Thanks @pfliu-nlp , looks good! I would suggest that you delete the dead code instead of commenting it out though.

PR #175, opened 2022-03-25
Added token-level eval metrics for condgen

neubig:
This adds some token-level metrics for conditional generation, similar to the span-level metrics for NER.

One thing that makes this slightly complicated is now there are different metrics included in a single report.
* Sentence-level metrics such as BLEU and ROUGE
* Token-level metrics f-measure, precision, and recall

This will probably have to be dealt with in the web front-end.

Note that training-set features are not exhaustively tested yet.

### DIFF
--- a/explainaboard/feature.py
+++ b/explainaboard/feature.py
@@ -599,9 +599,10 @@ class Features(dict):
     #     for k, v in dictionary.items():
     #         setattr(self,k,v)
 
-    def get_bucket_features(self) -> List:
+    def get_bucket_features(self, include_training_dependent=True) -> List:
         """
         Get features that would be used for bucketing
+        :param include_training_dependent: Include training-set dependent features
         :return:
         a list of features
         """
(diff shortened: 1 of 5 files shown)


pfliu-nlp:
Overall, it looks pretty good.
* this is the part that I probably forgot to delete in the previous PR, but we actually could remove it: https://github.com/neulab/ExplainaBoard/blob/4a67e9546df83675e91afe48d74be040bf50c9e1/explainaboard/processors/conditional_generation.py#L555

* >"One thing that makes this slightly complicated is now there are different metrics included in a single report." 

Yes, if front end could be in charge of this part, we then can remove the `for .. loop` part for the calculation of multiple metrics in SDK.

neubig:
Thanks a lot!

I actually have some ideas about restructuring the report file so that we can have multiple metrics calculated per bucket, which will be much more concise. But I'll save that for a future PR.

PR #176, opened 2022-03-25
Use classmethod in Processor for several properties.

odashi:
This pull request replaces some global properties in `Processor` to classmethods to bring better encapsulation.


neubig (approved):
LGTM! I merged main and added a unit test to check something I wasn't sure about, but this will be a nice change.

PR #179, opened 2022-03-26
Improve import consistency

odashi:
This change adds two flake8 plugins:

- `import-order`: constrain the order of import statements and newlining convention. I selected the google's style anyway. This improves readability of the statements and helps to find redundant imports.
- `absolute-import`: prohibits relative imports, meaning that the same module is imported by the same path from anywhere.

Additionally this change involves several fixes I found by overall flake8-ing.


neubig (approved):
LGTM, thanks!

PR #181, opened 2022-03-26
Fix CI scripts.

odashi:
This change adjusts the CI script for the current repository.
- Apply the newest version of each action.
- Utilize pre-commit action for invoking flake8 instead of manually installing.

Also fixes the setup.py:
- Add appropriate Python version tags -- 3.9 or later, since `sort_dict` uses the feature introduced by 3.9, and CI also uses 3.9.
- Remove black and flake8 dependencies. They are installed by pre-commit or manually in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,8 @@
 name: Continuous Integration
 on:
   push:
-    branches: ["*"]
+    branches:
+      - main
   pull_request:
     branches: ["*"]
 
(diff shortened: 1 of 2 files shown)


lyuyangh:
Thanks! Looks great! Because of the way I configured the triggers, I think a PR (like this) would trigger CI twice. I think we can fix this by restricting CI to only run on push to main + any PR. What do you think? Is there a better way to fix it?

lyuyangh (approved):

odashi:
@lyuyangh Yeah SGTM, we may need the on-push trigger for only the main branch.

odashi:
>  on-push trigger for only the main branch

Applied: 26f8de8

PR #182, opened 2022-03-26
Make the repository PEP-585 compliant.

odashi:
This change replaces several `typing` stuff which are deprecated in [PEP-585](https://peps.python.org/pep-0585/). The change also adds a pre-commit hook to apply the conversion automatically and an additional step in the lint CI to check that compliance.


neubig (approved):
LGTM! Thanks for all the code cleanup.

PR #183, opened 2022-03-27
Refactoring metrics

neubig:
This is starting to refactor metrics to save sufficient statistics and unifies the interface to evaluation. This speeds up and broadens support for significance tests/confidence intervals, speeds up calculation of bucketed scores, and reduces semi-duplicated code.

There's many things left to do:
* [x] Refactor HITs and MRR
* [x] Make EaaS-calculated metrics also follow the same framework
* ~~Implement harder metrics like BLEU~~ (this requires larger changes to EaaS as well, so I'm going to keep this for a future PR)
* [x] Implement metrics for QA
* ~~Implement metrics for NER~~ (this is a bit more complicated than the others, so I'll open an issue)
* [x] Add tests


explainaboard/metric.py
```diff
@@ -1,172 +1,373 @@
...
+
+        :param name: name of the metric
+        :param value: value of the metric
+        :param conf_interval: the confidence interval of the metric
+        :param conf_value: the p-value of the confidence interval
+        """
+        self.name = name
+        self.value = value
+        self.conf_interval = conf_interval
+        self.conf_value = conf_value
```
odashi, line 34: If these values are unchanged once they are initialized, would it be reasonable to use frozen dataclasses or tag them with the `Final` type hint? it is generally better to avoid exposing mutable members from usual classes.
If exposing the members is intended for some reason, could you provide details of the usage and appropriate type hints (since these members are typeless, users may insert any kinds of data after initializing the struct)?

ditto for other classes.

explainaboard/metric.py
```diff
@@ -1,172 +1,258 @@
...
+        self,
+        stats: np.ndarray,
+        conf_value: float,
+        n_samples: int = 2000,
+        prop_samples: float = 0.5,
+    ) -> tuple[float, float]:
+        """
+        :param stats: sufficient statistics as calculated by calc_stats_from_data
+        :param conf_value: the p-value of the interval
+        :param n_times: the number of bootstrapping samples
```
pfliu-nlp: should this be `n_samples` or `n_times`

explainaboard/metric.py
```diff
@@ -1,172 +1,258 @@
...
-        )
+    def calc_stats_from_data(self, true_data: list, pred_data: list) -> np.ndarray:
+        id_map = {}
+        for word in itertools.chain(true_data, pred_data):
+            if word not in id_map:
+                id_map[word] = len(id_map)
+        n_data = len(true_data)
+        n_classes = 3 * len(id_map)
+        # This is a bit memory inefficient if there's a large number of classes
+        stats = np.zeros((n_data, n_classes * 3))
```
pfliu-nlp, line 199: not sure if this place is worth adding some comments, such as:
*  `stats[i, id_map[t] * 3]` is used to store the number of actual samples in each class. 
*  `stats[i, id_map[p] * 3 + 1]` is used to store the number of predicted samples in each class.
*  `stats[i, id_map[t] * 3 + 2]`  denotes the number of correctly-predicted samples in class.

explainaboard/metric.py
```diff
@@ -1,172 +1,258 @@
...
+
+    def calc_metric_from_stats(self, stats: np.ndarray) -> float:
+        assert len(stats) % 3 == 0
+        n_classes = int(len(stats) / 3)
+        if self.average == 'micro':
+            match, true, pred = 0.0, 0.0, 0.0
+            for i in range(n_classes):
+                true += stats[i * 3]
+                pred += stats[i * 3 + 1]
+                match += stats[i * 3 + 1]
```
pfliu-nlp: should this be `match += stats[i * 3 + 2]`?

explainaboard/tests/test_aspect_based_sentiment_classification.py
```diff
@@ -12,7 +12,7 @@ class TestAspectBasedSentimentClassification(unittest.TestCase):
     def test_e2e(self):
         metadata = {
             "task_name": TaskType.aspect_based_sentiment_classification.value,
-            "metric_names": ["Accuracy", "F1score"],
+            "metric_names": ["Accuracy", "F1Score"],
```
pfliu-nlp, line 15: do we finally aim to replace this with metric's `default_name`?

explainaboard/metric.py
```diff
@@ -39,6 +39,23 @@ def to_dict(self):
...
 
 
+class MetricStats:
+    def __init__(self, data: np.ndarray):
+        self.data = data
+
+    def __len__(self):
+        return len(self.data)
+
+    def filter(self, indices: Union[list[int], np.ndarray]):
```
pfliu-nlp, line 58: Aha, I like this idea, nice preparation for the calculation of CI

explainaboard/metric.py
```diff
@@ -192,24 +192,24 @@ def calc_stats_from_data(self, true_data: list, pred_data: list) -> np.ndarray:
...
+        n_classes = int(len(agg_stats) / 3)
         if self.average == 'micro':
             match, true, pred = 0.0, 0.0, 0.0
             for i in range(n_classes):
-                true += stats[i * 3]
-                pred += stats[i * 3 + 1]
-                match += stats[i * 3 + 1]
+                true += agg_stats[i * 3]
+                pred += agg_stats[i * 3 + 1]
+                match += agg_stats[i * 3 + 1]
```
pfliu-nlp: mark

pfliu-nlp: I think this has been fixed in the latest commit.

explainaboard/metric.py
pfliu-nlp: I think this has been fixed in the latest commit.

explainaboard/metric.py
```diff
@@ -1,172 +1,373 @@
...
 
 import numpy as np
-import scipy
-import sklearn.metrics
-from sklearn.metrics import accuracy_score
+
+from explainaboard.utils.async_eaas import AsyncEaaSResult
+
+
+class MetricResult:
```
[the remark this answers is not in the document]
neubig: Thanks, done.

explainaboard/metric.py
neubig: Added it in the doc string for the whole function.

explainaboard/tests/test_aspect_based_sentiment_classification.py
neubig: I think thinking about this would be good, but not sure if it needs to be done now.

PR #184, opened 2022-03-27
Introduce isort instead of flake8-import-order

odashi:
This change introduces isort as the tool to adjust import order and disables flake8-import-order. This change involves a new pre-commit hook and corresponding CI.

@neubig please add the isort CI to the pull request checklist.


neubig (approved):
Thanks for the quick fix! This looks like it'll solve the problem.

PR #187, opened 2022-03-28
Applying mypy

odashi:
This change applies mypy (static type checker) to the whole repository and fixes potential bugs which are figured out by the tool.


explainaboard/loaders/kg_link_tail_prediction.py
```diff
@@ -29,7 +30,12 @@ def load(self) -> Iterable[dict]:
         :return: class object
         """
         data: list[dict] = []
-        raw_data = self.file_loaders[self._file_type].load_raw(self._data, self._source)
+
+        # TODO(odashi): Avoid potential bug: load_raw returns Iterable[Any] which is not a dict.
+        raw_data: dict[str, dict[str, str]] = self.file_loaders[  # type: ignore
+            unwrap(self._file_type)
+        ].load_raw(self._data, self._source)
```
odashi, line 37: @neubig please give me details about this.

explainaboard/processors/conditional_generation.py
```diff
@@ -427,6 +434,10 @@ def get_bucket_performance_tok(
                 toks_pred = samples_over_bucket_pred[bucket_interval]
 
             p_denom, r_denom = len(toks_pred), len(toks_true)
+
+            # TODO(odashi): I didn't understand what these lines are doing.
+            # These lines clearly violate type hints as far as I believed those in
+            # the argument list, but I couldn't figure out the correct typing.
```
odashi, line 440: @neubig please give me details about this.

explainaboard/utils/eval_ops.py
```diff
@@ -1,14 +1,14 @@
...
 from datalabs.operations.aggregate.auto_eval import auto_eval
 
 from explainaboard import FileType, get_loader, get_processor, Source
 
 
 @auto_eval(name="explainaboard")
-def explainaboard(samples: Iterator, dataset_info=None):
+def explainaboard(samples: str, dataset_info=None):
+    # TODO(odashi): I didn't understand what this docstring mean.
+    # This function seems to be not used anywhere, so I guessed the file can be removed.
```
odashi, line 11: @neubig please give me details about this.

odashi:
`test_ner` fails because `_gen_metric_stats` returns `None` which are to be unwrapped by the processor. As far as I saw the code this is expected, but it is unclear what is the correct fix.

explainaboard/loaders/kg_link_tail_prediction.py
lyuyangh: KgLinkTailPrediction is a special case I didn't think of. The system outputs are (key, value) pairs instead of a list of data points. `JSONFileLoader` uses `json.loads` which returns an any so I guess we can't guarantee it's an iterable anyways so maybe we should change the function signature of `load_raw()` to return any? What do you think?

explainaboard/utils/eval_ops.py
pfliu-nlp: This function could be removed.

explainaboard/config.py
```diff
@@ -1,13 +1,14 @@
 from dataclasses import dataclass
+from typing import Optional
 
 
 @dataclass
 class BuilderConfig:
```
neubig, line 6: I think this is unused code, it can just be deleted.

explainaboard/feature.py
```diff
@@ -333,54 +336,61 @@ def __post_init__(self):
...
+        return [self.int2str_scalar(v) for v in values]
+
+    @overload
+    def int2str(self, values: int) -> str:
+        ...
 
-        for v in values:
-            if not 0 <= v < self.num_classes:
-                raise ValueError("Invalid integer class label %d" % v)
+    @overload
```
neubig, line 384: Maybe I'm not familiar with this syntax, but just to confirm: is `int2str` intentionally defined twice or is this a typo?

explainaboard/info.py
```diff
@@ -47,23 +40,23 @@ class PaperInfo:
 
 @dataclass
 class Performance:
-    metric_name: str = None
```
neubig, line 50: `metric_name` and `value` are not actually optional.

explainaboard/info.py
```diff
@@ -47,23 +40,23 @@ class PaperInfo:
...
-    value: float = None
+    metric_name: Optional[str] = None
+    value: Optional[float] = None
     confidence_score_low: Optional[float] = None
     confidence_score_high: Optional[float] = None
 
 
 @dataclass
 class BucketPerformance(Performance):
-    bucket_name: str = None
```
neubig, line 58: `bucket_name` and `n_samples` are not actually optional.

explainaboard/info.py
```diff
@@ -99,7 +92,7 @@ class SysOutputInfo:
     # code: str = None
     # download_link: str = None
     # paper_info: PaperInfo = PaperInfo()
-    features: Features = None
```
neubig, line 102: I think `features` is not actually optional.

explainaboard/info.py
```diff
@@ -191,13 +184,13 @@ def copy(self) -> "SysOutputInfo":
 
 @dataclass
 class OverallStatistics:
-    sys_info: SysOutputInfo = None
```
neubig, line 194: I think `sys_info` is not actually optional.

explainaboard/info.py
```diff
@@ -191,13 +184,13 @@ def copy(self) -> "SysOutputInfo":
...
     metric_stats: Any = None
-    active_features: list[str] = None
-    overall_results: dict[str, Performance] = None
+    active_features: Optional[list[str]] = None
+    overall_results: Optional[dict[str, Performance]] = None
 
 
 @dataclass
 class FineGrainedStatistics:
-    samples_over_bucket: dict = None
```
neubig, line 202: I think these two are not actually optional.

explainaboard/loaders/file_loader.py
```diff
@@ -28,16 +28,13 @@ class FileLoaderField:
 
     src_name: Union[int, str]
     target_name: str
-    dtype: Union[type[int], type[float], type[str]] = None
```
neubig, line 31: remove = None

explainaboard/loaders/kg_link_tail_prediction.py
neubig: Yeah, I think `load_raw` loads raw data, so we shouldn't be over-prescriptive of its format.

Also, maybe we should somehow indicate that `load_raw` shouldn't be used outside of `explainaboard.loaders`?

explainaboard/metric.py
```diff
@@ -56,13 +57,13 @@ def __len__(self) -> int:
         """
         Returns the number of samples in the dataset
         """
-        return len(self._data)
```
neubig, line 59: This is actually a bug, fixed in https://github.com/neulab/ExplainaBoard/pull/188

explainaboard/metric.py
```diff
@@ -317,6 +318,9 @@ def __init__(self, name: str, eaas_result: AsyncEaaSResult):
         super().__init__(data=None)
         self.name = name
         self.eaas_result = eaas_result
+        self._data: Optional[np.ndarray] = None
+
+        # TODO(odashi): remove this field: this is private but unused.
         self._corpus_value = None
```
neubig, line 324: This is returned by `_get_corpus_value()` and may be useful in the future.

explainaboard/metric.py
```diff
@@ -334,18 +338,18 @@ def get_corpus_value(self) -> float:
         Return the evaluation metric value over all examples in the corpus.
         """
         self._fetch_results()
-        return self._corpus_value
+        return unwrap(self._corpus_value)
```
neubig, line 341: This unwrap is probably not necessary, as `_fetch_results()` guarantees that it will be initialized. But it probably won't hurt to have it here.

explainaboard/processors/conditional_generation.py
```diff
@@ -211,23 +213,24 @@ def _gen_metric_stats(self, sys_info: SysOutputInfo, sys_output: list[dict]) ->
...
 
-    def _fetch_metric_stats(self, metric_stats: Any):
+    # TODO(odashi): Restructure this function (and EaaS client) to be type-safe.
+    def _fetch_metric_stats(self, metric_stats: dict[str, Any]):
         """
         A utility function used to lazily fetch the actual scoring dict when it's necessary.
         """
         if 'request_id' in metric_stats:
-            eaas_stats = self._eaas_client.wait_and_get_result(
+            eaas_stats: dict[str, Any] = unwrap(self._eaas_client).wait_and_get_result(
```
neubig, line 233: could be `self.get_eaas_client()` instead.

explainaboard/processors/conditional_generation.py
neubig: The four dimensional data structure is saying:
* What sentence is it? `x[0]`
* What variety of information about the sentence? `hyp_tok_info`: information about the hypothesis tokens.
* What token is it? `x[1]`
* What variety of information about they hypothesis token? `tok_matched`: whether or not the token matches a token in the reference.

And analogously for the reference tokens, whether they match the hypothesis.

neubig (commented):
Thanks a lot for this!

I think `sys_info.metric_names` and `sys_info.features` could probably be listed as not optional, which would allow for many of the `unwrap`s to be removed.

explainaboard/processors/conditional_generation.py
odashi: so does the code here intend `dict[(int, str, int, str) -> something]`? as `x` has the type of `str`, the first and third subscription is warned as type mismatch. The type information here is inherited from the function signature and I couldn't figure out any quick fix on this function.

explainaboard/metric.py
neubig: Sure, sounds good.

explainaboard/feature.py
neubig: OK, thanks for the clarification.

pfliu-nlp:
(if possible, please don't delete this branch even after the merge, I will keep it for the learning purpose.)

explainaboard/loaders/kg_link_tail_prediction.py
neubig: Sounds good. Feel free to either do this yourself in this PR or assign an issue to me for completion after this PR is merged.

explainaboard/metric.py
neubig: I'd like to clarify the issue here:
1. `self._corpus_value` is set to `None` in the constructor (because this class does lazy evaluation of statistics, it is not initially set)
2. it is set to an appropriate value in in `_fetch_results()` (when lazy evaluation actually occurs)
3. it is returned by `get_corpus_value()`

This seems like a pretty standard way to implement lazy evaluation, but perhaps it could be documented better? I'm happy to do that, but I just wanted to clarify the issue.

Or is the issue that `get_corpus_value()` is not called anywhere else in the codebase? If that's the issue, I think that it's still reasonable to have it as an accessor, given that the EaaS client returns this value and it may be useful in the future.

odashi: I read the impl. again and understood that `_fetch_results` calls an external invocation (EaaS) which may take long time, which is documented on the docstring but my editor didn't show the rightmost part of the sentence-_- Thanks for clarifying.
Lazy evaluation is actually achieved by flagging `not self._data` for every members. This is double-purposing a variable and introducing a boolean flag with explicit name (e.g., `_initialized`) is recommended.

explainaboard/processors/conditional_generation.py
odashi: @neubig Could you separate the appropriate part into another pr? I can try to merge it anyway but it requires certain cherry-picking maybe

explainaboard/loaders/kg_link_tail_prediction.py
odashi: Keeps the todo comment and resolves this conversation.

explainaboard/utils/eval_ops.py
odashi: Removed: 17309a5

explainaboard/config.py
odashi: Removed: 3009a3e

explainaboard/info.py
odashi: Fixed: bc84d7b

explainaboard/info.py
odashi: This is not allowed by [PEP-557](https://peps.python.org/pep-0557/#specification) because:
- The members in `BucketPerformance` follows all the members in `Performance`
- Attributes without defaults (`bucket_name` here) could not follow attributes with defaults (`confidence_score_high` here)

odashi: Subclassing dataclasses is usually not a good choice because of this kind of problems. In this case owning `Performance` by `BucketPerformance` is the correct way IMO:

```python
@dataclass
def BucketPerformance:
  ...members...
  performance: Performance
```

explainaboard/info.py
odashi: Fixed: ba85ab9

explainaboard/info.py
odashi: Fixed: 4c127d6

explainaboard/metric.py
odashi: Keeps the change as-is for avoiding the typing error. This should be resolved at merging #188, or separate pr to resolve this specific problem is needed.

explainaboard/processors/conditional_generation.py
neubig: let's merge this mypy PR and I'll fix the merge errors later

neubig (approved):

Issue #161, opened 2022-03-22
Conditional generation bucket-level evaluation should use proper statistics


neubig:
Currently conditional generation bucket-level evaluation is taking the average of segment scores, which only works if a metric is decomposable across sentences. This is not the case for major metrics like BLEU and ROUGE, so this should be fixed.

pfliu-nlp:
Good capture.
>"This is not the case for major metrics like BLEU and ROUGE, so this should be fixed"
I think ROUGE  should be decomposable.  
So far, what eaas supports:
* regard bleu chrf as non-decomposable metrics and calculate sample-level (without sufficient statistics) and dataset-level scores
* regard other metrics as decomposable metrics and calculate the sample-level (without sufficient statistics) and dataset-level 
Maybe a good way to achieve this is that eass only need to return sample-level results together with sufficient statistics, and explainaboard could aggregate them.

neubig:
Sorry, to be clear, by decomposable I mean this:

`metric(Y,Y*) = 1/|Y| \sum_{(y,y*) \in (Y,Y*)} metric(y,y*)`

This is not the case for BLEU, ROUGE, chrF. But I'm pretty sure it is the case for COMET, BERTScore, BARTScore, etc.

All of the above metrics can be calculated from the sum of sufficient statistics, which is what we should do. I'll work on this unless someone else wants to work on it.

neubig:
So in the [original definition of ROUGE-N](https://aclanthology.org/W04-1013.pdf), it is not the case, as shown by this equation:
![Screen Shot 2022-03-22 at 4 04 22 PM](https://user-images.githubusercontent.com/398875/159566434-cce55580-6788-45be-9ab9-ec01e1030408.png)

But I confirmed that indeed the Google toolkit for example just takes the average of sample-level scores.

PR #188, opened 2022-03-28
Added tests for EaaS-based metrics

neubig:
This PR adds a test that fails due to this bug:
https://github.com/neulab/ExplainaBoard/issues/161

Fixing the bug will take a certain amount of time, but I think that probably this PR should be incorporated (with failing tests), and then we should aim to fix the code so the tests pass ASAP. (Also, this fixes some small bugs in the metrics package)


explainaboard/tests/test_metric.py
```diff
@@ -43,3 +54,81 @@ def test_mrr(self):
...
+        # Initialize client and decide which metrics to test
+        eaas_client = AsyncEaaSClient()
+        eaas_client.load_config(Config())
+        metric_names = ['rouge1', 'bleu', 'chrf']
+        # Uncomment the following line to test all metrics, but beware that it will be very slow
+        # metric_names = eaas_client._valid_metrics
+        metrics = [explainaboard.metric.EaaSMetric(name=name) for name in metric_names]
+
+        # Get results for full data and half data
+        half_bound = int(len(sys_output))
```
pfliu-nlp, line 103: Is this the expected value of `half_bound`?

pfliu-nlp (approved):

PR #192, opened 2022-03-30
Add confidence interval calc to CLI

neubig:
This
1. adds an option for the p-value of confidence interval calculation to the CLI
2. sets the default of confidence interval calculation to 0.05 so confidence intervals are calculated by default


pfliu-nlp (approved):
looks good!

PR #194, opened 2022-03-31
Exploring a good way for users to perform multiple system analyses

pfliu-nlp:
This PR mainly attempts to explore a good way for users to perform multiple system analyses, which would be useful for several applications:

(1) benchmark evaluation
    * how to submit
    * analyze (aggregating results from multiple tasks)
(2) multi-lingual evaluation
   * how to submit
   * how to analyze
(3) System combination
(4) Comparative analysis over two batches of systems

The major modifications this PR made are
* Relatively generalize the customized feature functionality by introducing  `user_defined_features_configs` in the Loader class (We can skip KG task, whose format would be re-designed).
* Introduce the concept of `user_defined_metadata_configs`, which could be optionally included in the system output file (I did some surveys from students, and the conclusion is this is acceptable.)
To give one example for text classification task:
image

* Support getting metadata from the system output file instead of command lines.
To give one example, 
```
explainaboard --system_outputs ./data/system_outputs/multilingual/json/CL-mt5base/marc/test-de_9335.json 
```
then,
image

* Add figure drawing module in SDK (explainaboard/analyzer), so far, histograms could be generated given a system output.
For example (figures will be further refined),
image

* CLI supports the evaluation of a batch of system outputs. Two folders of files would be automatically generated: (a) analysis report files,  (2) beautiful histograms of each analysis report so that users could use them for their papers.
To give one example,

```
explainaboard --system_outputs ./data/system_outputs/multilingual/json/CL-mt5base/xnli/*
```

* Because of the above support, the current ExplainaBoard SDK roughly supports  Multilingual Multitask Evaluation. This put forward a requirement for a new type of fine-grained analysis in terms of languages, datasets, models (dubbed meta-level fine-grained analysis for example).

* Support meta-level fine-grained analysis for multilingual multitask evaluation, which basically involves two steps:
    * Generating all reports based on a collection of system outputs from different languages and different tasks
    * Performing different types of aggregations based on the above-generated reports and generating useful statistics

To give one example, 
```
explainaboard --reports ./output/reports/*
```
Then

image

* Support simple visualization in the command line for meta-level fine-grained analysis

To give one example, 

```
explainaboard --reports ./output/reports/* --datasets xnli marc
```
image

```
explainaboard --reports ./output/reports/* --datasets xnli marc --models_aggregation minus

```

image

I think there are many things worth discussing here, and this is a concept proving implementation. I will make this PR ready to review once I have finished the necessary documentation and formating fixing.


explainaboard/info.py
```diff
@@ -87,6 +87,8 @@ class SysOutputInfo:
     metric_names: Optional[list[str]] = None
     reload_stat: bool = True
     is_print_case: bool = True
+    is_print_confidence_interval: bool = False
```
neubig, line 90: `is_print_confidence_interval` has been replaced by `conf_value`. Also, I think confidence intervals should be on by default, no reason to have them off really.

neubig (approved):
I reviewed this and I think it's good progress in the right direction! I personally would be interested in seeing it incorporated sooner rather than later so I could also use some of the functionality here. We could refactor/increase documentation later as necessary.

A few overall suggestions:
* `user_defined_metadata_configs` is very verbose. Maybe just `metadata` is fine, especially for the stuff we put in the files.
* Maybe some of the aggregation stuff could be moved outside of `explainaboard_main.py` into a separate file?
* Should `analyzers` be called `visualizers`?

explainaboard/info.py
pfliu-nlp: >"user_defined_metadata_configs is very verbose. Maybe just metadata is fine"

Good point. So far I used it to make a consistent naming with `user_defined_features_configs`, We can re-organize them together later. Also some new and unexpected `xyz-config` may come in the future.

> "is_print_confidence_interval has been replaced"

OK. removed!

>"Should analyzers be called visualizers"

I have considered it, so far I reuse the folder `analyzer` that Chihao has set up.

docs/multilingual_multitask_evaluation.md
```diff
@@ -0,0 +1,255 @@
+# Multilingual Multitask Evaluation
```
neubig, line 1: "Multilingual/Multitask" might be better to indicate that it's "and/or" not "and".

Similarly below, "multilingual and/or multitask evaluation" might be good.

neubig (commented):
Looks great! Just a few small comments.

Could we do a global search and replace for `user_defined_metatdata_configs` and replace it with something like `metadata` or `config_metadata`? The current variable name seems too verbose and is not adding much.

explainaboard/info.py
```diff
@@ -88,8 +88,9 @@ class SysOutputInfo:
     metric_names: Optional[list[str]] = None
     reload_stat: bool = True
     is_print_case: bool = True
+    language: str = "en"
```
neubig, line 91: Seems "language" is included twice.

PR #195, opened 2022-03-31
Complying with E501

odashi:
This change revives the E501 Flake8 option (line too long).

There is a large diff because of separating `BASIC_WORDS` into another file.

- [x] This change should be blocked until #196 has been merged.


odashi:
Black is also failing due to well-known bug: https://github.com/psf/black/issues/2964
we need upgrading it

odashi:
#196 fixes the issue around Black.

neubig (approved):
LGTM, thanks!

PR #196, opened 2022-03-31
Use pre-commit Black in CI.

odashi:
This change swaps the Black CI binary to that of pre-commit to use the same config.


neubig (approved):
Thanks!

PR #198, opened 2022-03-31
update metric.py

pfliu-nlp:


explainaboard/metric.py
```diff
@@ -418,12 +418,21 @@ class QAMetric(Metric):
 
     def normalize_answer(self, s: str) -> str:
         """Lower text and remove punctuation, articles and extra whitespace."""
-        s = re.sub(r'\b(a|an|the)\b', ' ', s)
-        s = ' '.join(s.split())
-        exclude_punc = set(string.punctuation)
-        s = ''.join(ch for ch in s if ch not in exclude_punc)
-        s = s.lower()
-        return s
+
```
neubig, line 421: Sorry, I realize my bug now. I'd suggest the following code, which should be equivalent to the previous code but less verbose:

        s = s.lower()
        exclude_punc = set(string.punctuation)
        s = ''.join(ch for ch in s if ch not in exclude_punc)
        s = re.sub(r'\b(a|an|the)\b', ' ', s)
        s = ' '.join(s.split())
        return s

pfliu-nlp: Hey, Graham, thanks!  I just introduced the concept of language to the Metric class.  So far, I followed the official script (from MLQA) Jinlan shared with me (also a little verbose). You probably could take a look.

neubig: The following look a bit specific to the MLQA dataset and seems like maybe we shouldn't have it directly in the top-level QAMetric class?

As an alternative suggestion, maybe we could have an analogous class to `Tokenizer` that is something like `Preprocessor`, and we could have `MLQAPreprocessor`. Then `QAMetric` could take a `Preprocessor` as an argument, and apply the appropriate processing in a more generic way before doing actual evaluation.

explainaboard/tests/test_extractive_qa.py
```diff
@@ -33,6 +34,10 @@ def test_generate_system_analysis(self):
         # analysis.write_to_directory("./")
         self.assertIsNotNone(sys_info.results.fine_grained)
         self.assertGreater(len(sys_info.results.overall), 0)
+        # should be 0.6974789915966386
+        self.assertGreater(sys_info.results.overall["ExactMatchQA"].value, 0.69)
```
neubig, line 38: Maybe `assertAlmostEqual` is more appropriate?

pfliu-nlp: >"As an alternative suggestion, maybe w..."

I actually like this idea and will try to go towards this end.


>"Maybe assertAlmostEqual is more appropriate?"

Interesting! it seems that python has provided everything...

pfliu-nlp:
@neubig  this should be ready for review.

explainaboard/metric.py
```diff
@@ -91,12 +90,16 @@ def default_name(cls) -> str:
         """Returns the default name of the metric."""
         ...
 
-    def __init__(self, name: str = None):
+    def __init__(
```
neubig, line 93: Suggestion: could we replace these individual variables with a class `MetricConfig`? The reason why I suggest this is that then the MetricConfig class could be (1) passed directly to the various metric computing functions, and (2) saved in a database or report, allowing the evaluation setup to be reproducible.

explainaboard/metric.py
```diff
@@ -218,6 +221,8 @@ def default_name(cls) -> str:
 
     def __init__(
         self,
+        language="en",
```
neubig, line 224: I think these two lines should probably be removed, as it's essentially defining the defaults in two places, which might cause problems later if you change the defaults somewhere.

explainaboard/utils/preprocessor.py
```diff
@@ -0,0 +1,84 @@
...
+import re
+import string
+import sys
+import unicodedata
+
+from explainaboard.utils.tokenizer import MLQAMixTokenizer, SingleSpaceTokenizer
+
+
+class Preprocessor:
+    def __call__(self, text: str) -> str:
```
neubig, line 12: Add `@abc.abstractmethod` decorator, and replace `raise NotImplementedError` with `...`

explainaboard/utils/preprocessor.py
```diff
@@ -0,0 +1,84 @@
...
+            return ''.join(ch for ch in text if ch not in self.PUNCT)
+
+        def lower(text):
+            return text.lower()
+
+        return white_space_fix(
+            remove_articles(remove_punc(lower(s)), language), language
+        )
+
+    def __call__(self, text: str, language: str) -> str:
```
neubig, line 83: This adds another variable to the abstract `__call__` function of `Preprocessor`. I'd suggest specifying the language in the `__init__` function instead, and removing it from the arguments here.

explainaboard/metric.py
```diff
@@ -523,7 +523,8 @@ def default_name(cls) -> str:
     def sample_level_metric(self, ground_truth: str, prediction: str) -> float:
         return (
             1.0
-            if self.normalize_answer(prediction) == self.normalize_answer(ground_truth)
+            if self.preprocessor(prediction, self.language)
```
neubig, line 526: Per comments above, probably better not to specify `self.language` here.

explainaboard/metric.py
```diff
@@ -538,8 +539,8 @@ def default_name(cls) -> str:
         return 'F1ScoreQA'
 
     def sample_level_metric(self, ground_truth: str, prediction: str):
-        prediction_tokens = self.normalize_answer(prediction).split()
-        ground_truth_tokens = self.normalize_answer(ground_truth).split()
+        prediction_tokens = self.preprocessor(prediction, self.language).split()
```
neubig, line 542: Similarly.

explainaboard/processors/extractive_qa.py
```diff
@@ -10,10 +10,15 @@
...
 from explainaboard.tasks import TaskType
 import explainaboard.utils.feature_funcs
+from explainaboard.utils.preprocessor import MLQAPreprocessor
 
 
 @register_processor(TaskType.question_answering_extractive)
 class QAExtractiveProcessor(Processor):
+    def __init__(self):
+        super().__init__()
+        self._preprocessor: MLQAPreprocessor = MLQAPreprocessor()
```
neubig, line 20: Processors should not include data/task-specific information in their member variables (to make sure they remain "stateless"), but the type of preprocessing performed might be task-specific. I think it'd be better for this to be stored somewhere else, maybe having `MetricConfig` included in the `SysInfo` class for example.

explainaboard/tests/test_extractive_qa.py
neubig: This still needs to be fixed I guess.

neubig (changes requested):
Thanks! This is looking better, but I left some suggestions.

explainaboard/metric.py
pfliu-nlp: Sounds good, I will make it an issue for another future PR.

explainaboard/metric.py
pfliu-nlp: It will lead to an error if these two are removed due to: https://github.com/neulab/ExplainaBoard/blob/35d7af9b858ea7901038333f40581728a9f49e29/explainaboard/processors/processor.py#L142

explainaboard/utils/preprocessor.py
pfliu-nlp: sounds good, do we want to do a similar thing for the `tokenizer` class?

explainaboard/utils/preprocessor.py
pfliu-nlp: I hesitate about this for a while. 
>"I'd suggest specifying the language in the __init__ function instead"
What's the advantage of doing this?

explainaboard/processors/extractive_qa.py
pfliu-nlp: For future PR

pfliu-nlp:
Modified part of them. Others will be put in future PR.

explainaboard/utils/preprocessor.py
neubig: Yeah, good idea. I just recently learned this syntax from @odashi :)

explainaboard/utils/preprocessor.py
neubig: The reason why is that other preprocessors might have other variables than `language` that specify their behavior. For example, they might have to specify a model, or a regex, or something else. If we specify it in the `__call__` function, then we'll have to have different behavior when the preprocessor is used, so the call would have to be modified (in every place where the processor is used).

explainaboard/metric.py
neubig: OK, we can fix this in a separate PR when we introduce MetricConfig.

neubig:
I think that this is OK, but I would really like to keep the interface of the `__call__` function of `Preprocessor` as simple as possible (i.e. take only the input string as an argument). If you can just fix that part I think this is ready to merge.

pfliu-nlp:
> I think that this is OK, but I would really like to keep the interface of the `__call__` function of `Preprocessor` as simple as possible (i.e. take only the input string as an argument). If you can just fix that part I think this is ready to merge.

Sounds good, This has been fixed in this pr: https://github.com/neulab/ExplainaBoard/pull/194

Issue #190, opened 2022-03-30
NER should be converted to use Metric class


neubig:
Currently only the NER task is not using a standard Metric class which is inconvenient, because it makes the interface different and also prevents the easy application of significance tests. It's also causing one failing unit test, and should be fixed ASAP.

PR #202, opened 2022-04-02
Change NER to use metrics class

neubig:
Fixes https://github.com/neulab/ExplainaBoard/issues/190


explainaboard/utils/span_utils.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import annotations
+
+
+def get_spans_from_bio(seq):
```
pfliu-nlp, line 4: I actually haven't figured out the content of `seq`  especially after reading the docstring, should it be list[str], such as ["O", "B-PER", "I-PER"]?

pfliu-nlp (approved):
Overall, it looks nice to me.

explainaboard/utils/span_utils.py
neubig: Thanks! This doc was legacy doc from before, but I made it clearer (and also added a test checking the desired behavior)

pfliu-nlp: nice! Feel free to merge this PR!

neubig:
Thanks for the review! I also bumped the version and fixed the EaaS interface to use the latest version of the client. This will be release 0.8.5

Issue #197, opened 2022-03-31
Bug(?): the evaluation results using `ExactMatchQA` on extractive qa task


pfliu-nlp:
Reported from Jinlan: the evaluation results using `ExactMatchQA` on the extractive QA task are significantly lower than the one calculated by the previous version of SDK

pfliu-nlp:
Aha, thanks Graham, I'm looking now, by comparing with the old branch, I found the `normalize_answer` function has been re-written, but I think they are not completely the same, is this intended?

neubig:
No, they were supposed to be the same. If it's not that's my mistake. Thanks for fixing this and sorry for the bug!

neubig:
Being handled by this PR: https://github.com/neulab/ExplainaBoard/pull/198

PR #205, opened 2022-04-02
Fixed bug in QA metrics

neubig:
Fixes https://github.com/neulab/ExplainaBoard/issues/197

Note that @pfliu-nlp also has a PR open (https://github.com/neulab/ExplainaBoard/pull/198) that also includes the bigger change of doing multilingual evaluation metrics. I think this is a good PR, but it might take a little bit of design to get it right, so I'd suggest making this smaller change first so we can do a release of the SDK and start working on the web interface again.


pfliu-nlp (approved):
>" I'd suggest making this smaller change first so we can do a release of the SDK and start working on the web interface again."

Sounds good and agree.

PR #214, opened 2022-04-06
Support dynamic hits k

pfliu-nlp:
This is a lightweight enhancement, to make the `Hits` metric support dynamic `k`.

### DIFF
--- a/explainaboard/analyzers/bar_chart.py
+++ b/explainaboard/analyzers/bar_chart.py
@@ -1,5 +1,7 @@
+from __future__ import annotations
+
 import logging
-from typing import Optional
+from typing import Any, Optional
 
 from matplotlib import pyplot as plt
 
@@ -9,13 +11,13 @@ mlogger = logging.getLogger('matplotlib')
 mlogger.setLevel(logging.WARNING)
 
 
-def mp_format(data) -> dict:
+def mp_format(data: list[dict[str, Any]]) -> dict:
     """
     Adapt the format of data
     :param data:
     :return:
     """
-    data_mp = {
+    data_mp: dict = {
         "x": [],
         "y": [],
         "y_errormin": [],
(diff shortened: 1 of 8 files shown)


explainaboard/analyzers/bar_chart.py
```diff
@@ -107,17 +107,18 @@ def plot(
...
-            data_mp["x"],
-            data_mp["y"],
-            yerr=[data_mp["y_errormin"], data_mp["y_errormax"]],
-            elinewidth=5,
-            capsize=5,
-        )  # you can use color ="r" for red or skip to default as blue
-        for cap in caps:
-            cap.set_color('gray')
-            cap.set_markeredgewidth(10)
+    # if data_mp["y_errormin"][0] is not None:
```
neubig, line 110: Delete code instead of commenting.

explainaboard/processors/kg_link_tail_prediction.py
```diff
@@ -111,6 +113,17 @@ def default_features(cls) -> feature.Features:
     def default_metrics(cls) -> list[str]:
         return ["Hits", "MeanReciprocalRank"]
 
+    def _get_metrics(
+        self, sys_info: SysOutputInfo
+    ) -> list[explainaboard.metric.Metric]:
+        config = explainaboard.metric.HitsConfig(
+            hits_k=5
```
neubig, line 120: Two comments:

Shall we do the following instead?
`config = HitsConfig(...) if name == 'Hits' else MetricConfig()`

5 is already the default elsewhere, so we probably shouldn't set it here. If it is set here, then even if we change the default in `metric.py` the resulting value would not change, which could be confusing.

neubig (commented):
Looks good, I approve with three small suggested changes.

explainaboard/processors/kg_link_tail_prediction.py
pfliu-nlp: sounds good!

pfliu-nlp:
Merge this first and will fix one unit test error in another PR.

PR #222, opened 2022-04-09
Increase EAAS dependency, bump version

neubig:
There was an issue with our deployment pipeline for `eaas_client` that caused versions to be deployed too early, so 0.3.7 was broken. I fixed this, deployed a new version 0.3.8, and increased the required version of eaas, so hopefully this PR will pass tests.

For good measure I also bumped the version of explainaboard.


pfliu-nlp (approved):
looks good

PR #223, opened 2022-04-09
refactor: move custom feature config parsing logic to base Loader

lyuyangh:
- FileLoader is designed to only have knowledge of the file type so I moved the custom feature parsing logic to the base Loader. It will parse the config dict and add custom fields to the fields list. Then, `JSONFileLoader` can load these custom features as if they are the default ones.
- Defined a dataclass for `CustomFeature` in loader.py. I'm not sure if it's the best place so please let me know if you want me to move it.
- Added type hints for https://github.com/neulab/ExplainaBoard/blob/27c30fa06f77883a4fe92d8fb778e1b79e874342/explainaboard/processors/processor.py#L257 just so it passes mypy. `_complete_features` is quite complicated, I think it maybe helpful if we can refactor some of the logic in there when we get a chance.

### DIFF
--- a/explainaboard/loaders/file_loader.py
+++ b/explainaboard/loaders/file_loader.py
@@ -5,7 +5,7 @@ import csv
 from dataclasses import dataclass
 from io import StringIO
 import json
-from typing import Any, Optional, Union
+from typing import Any, final, Optional, Union
 
 from explainaboard.constants import Source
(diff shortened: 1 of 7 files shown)


neubig (approved):
LGTM, seems like this is basically housekeeping

PR #225, opened 2022-04-10
Small bug fixes: TSV loading + required system outputs

neubig:
Just some fixes to small bugs I encountered when trying to use the software.
* I believe system outputs should be required because ExplainaBoard can't work without them (and throws an opaque error)
* I removed quoting in the TSV loader, as it through an opaque error if there was a TSV file that contained quotes.


lyuyangh (approved):
Looks good! Thanks!

PR #226, opened 2022-04-10
Added more options for tokenization

neubig:
This:
1. Makes the tokenizer a property of `sys_info` as opposed to `processor`
2. Adds the ability to switch default tokenizers based on task and language
3. Adds the SacreBLEU tokenizers, and makes them default for conditional generation analysis


pfliu-nlp (approved):
Look good to me. I guess in the future, more and more config-based variables will be moved to  `SysOutputInfo` class to achieve more automated and customized features, such as
* preprocessor
* metric config
* bucketing config (maybe)

PR #229, opened 2022-04-11
Feat: Separate sys output

### LINKED ISSUE #141: Separate dataset and system output

lyuyangh:
closes #141 
This is a major refactoring of the loaders to split the dataset (features+true label) and the system output (predicted label). This is a major change to our loader API so please let me know if there's anything you want me to change.
1. There are now two `get_loader()` variants:
    1. `get_loader()` loads from datalab. The user needs to specify dataset name, subset name (optional), and split (defaults to "test"). Only text classification and extractive_qa have `DatalabFileLoader` implemented because those are the only ones that were supported in this SDK. It should be easy to implement this for all the other tasks. I'll check data lab's documentation and implement them next.
    2. `get_loader_custom_dataset()` loads from a custom dataset file provided by the user
2. I updated all the unit tests and made sure they pass but I only tested `get_loader()` on sst2 dataset. It'll take more time to test this feature on other datasets because I need to generate fake system output files. I'll continue to do more tests. I also made several test sys output files smaller so the tests run faster and it'll be easier for us to write tests.
3. I also updated `explainaboard_main.py` but I was only able to test the following commands (the ones in README). main.py has many parameters and it's quite complicated. I think it may be easier for us to maintain it if we split the script into several scripts (e.g. one for analyzing reports, one for pairwise analysis, one for multiple systems, etc.)
```
# datalab dataset
explainaboard --task text-classification --system_outputs ./data/system_outputs/sst2/sst2-lstm-output.txt --dataset sst2

# custom dataset
explainaboard --task text-classification --system_outputs ./data/system_outputs/sst2/sst2-lstm-output.txt --custom_dataset_paths ./data/system_outputs/sst2/sst2-dataset.tsv
```
4. I updated the README. I think several examples in the README are taken directly from the tests. Do you think we should point/link users to the tests in the README instead (so we don't have to update the README if we change the tests)?
5. I added a `narrow()` method in **typing_utils.py** that does type narrowing for you with `isinstance()`. This may be a safer (but less efficient) alternative to typecasting?
6. There are many more system output files in the repo that needs to be updated. I'll work on this next.


explainaboard/loaders/loader.py
```diff
@@ -87,78 +128,66 @@ def user_defined_metadata_configs(self) -> dict:
             )
         return self._user_defined_metadata_configs
 
-    def _parse_user_defined_features_configs(self) -> dict:
-        if self._file_type == FileType.json:
-            raw_data = self.file_loaders[FileType.json].load_raw(
-                self._data, self._source
+    def _parse_user_defined_fields(self) -> tuple[dict[str, CustomFeature], dict]:
+        """custom features and metadata can only be defined in the output file and it
```
neubig, line 132: This is something that might need discussion. There are certainly instances where we would want the user-defined fields within the dataset. For example, the MetaphorQA dataset, for which we're planning to build a leaderboard, has annotations over the type of questions included in the dataset itself.

On the other hand, ExplainaBoard users might want to define their own features to use in analyses, in which case they'd want to upload them either as part of the system output, or as a separate file.

I'm not sure if this is important enough to block this PR, but it's worth thinking about what our future strategy will be.

explainaboard/loaders/loader.py
```diff
@@ -87,78 +128,66 @@ def user_defined_metadata_configs(self) -> dict:
...
-    return register_loader_fn
+    def load(self) -> list[dict]:
+        dataset_loaded_data = self._dataset_file_loader.load(
+            self._dataset_data, self._dataset_source
+        )
+        output_loaded_data = self._output_file_loader.load(
+            self._output_data, self._output_source
+        )
+        if len(dataset_loaded_data) != len(output_loaded_data):
+            raise Exception(
```
neubig, line 184: Maybe `ValueError`? I try to avoid just `Exception`.

explainaboard/loaders/named_entity_recognition.py
```diff
@@ -14,5 +19,25 @@ class NERLoader(Loader):
...
+                [
+                    FileLoaderField(0, "tokens", str),
+                    FileLoaderField(1, "true_tags", str),
+                ]
+            )
+        }
+
+    @classmethod
+    def default_output_file_loaders(cls) -> dict[FileType, FileLoader]:
+        return {FileType.conll: CoNLLFileLoader([FileLoaderField(0, "pred_tags", str)])}
```
neubig, line 43: For CoNLL format it'd be a bit weird to just output the tags (very non-standard), so I think we should probably assume that it'll still have the tokens in the first column.

neubig (approved):
Looks great, thanks for this absolutely massive commit!

I think the changes that I requested below are relatively minor, which is why I'm hitting "approve", but please do consider them.

explainaboard/loaders/loader.py
lyuyangh: I see. I think putting `user_defined_features_configs` and `user_defined_metadata_configs` in the dataset or output file can be a little awkward in some cases because it would mean that the file format can be different depending on whether user-defined configs exist. It also makes the JSON format a special case because user-defined fields can only be specified with this file type.

Also, instead of passing in the task name explicitly, the SDK now supports passing in the task names in the metadata config. But that means that we need to parse the file before knowing which loader we need to parse it with (loaders are task-dependent). Currently, we use dummy task types to parse the file once to get the metadata and then create an actually task-specific loader to parse the file a second time. (this is in **explainaboard_main.py**).

I think maybe it would make sense to move user-defined configs to a separate file but this would require our users to provide one more file. If we move these configs to a separate file, I think we can allow configs and metadata to come from two sources: a config file (maybe yaml or json) and the metadata argument passed to `get_loader` or `get_processor` in the code. We will merge the two to get the full metadata. I think part of this is already implemented in **explainaboard_main.py** as a special case where the task names are extracted from the output file and added to the metadata dict. We can maybe support this feature for all general cases.

explainaboard/loaders/loader.py
lyuyangh: Actually, this validation isn't relevant anymore. `self._user_defined_features_configs` and `self._user_defined_metadata_configs` are always of dict types (this is enforced by type hints) so I have removed this.

explainaboard/tests/artifacts/ner/dataset.tsv
```diff
@@ -1,6 +1,6 @@
-SOCCER	O	O
--	O	O
-JAPAN	B-LOC	B-MISC
+SOCCER	O
```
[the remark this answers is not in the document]
lyuyangh: I removed the extra predicted label column from the file.

explainaboard/loaders/named_entity_recognition.py
lyuyangh: Got it! I added the token column back to the output file for CoNLL.

docs/add_new_formats.md
```diff
@@ -20,13 +20,20 @@ For example, suppose that you have a system output of the summarization task
 in `tsv` format:
 
 ```python
-from explainaboard import TaskType, get_loader, get_processor
-
-path_data = "./artifacts/test-summ.tsv"
-loader = get_loader(task = TaskType.summarization, 
-                    file_type = FileType.tsv,
-                    data = path_data)
+from explainaboard import TaskType, get_loader, get_loader_custom_dataset
```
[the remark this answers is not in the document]
neubig: Ahh, sorry, I didn't read the explanation in the original PR closely enough and was confused about these two.

What about
* `get_standard_dataset_loader` or `get_datalab_loader`
* `get_custom_dataset_loader`
to make the contrast clear?

neubig:
> I changed return types of FileLoader.load() and FileLoader.load_raw() from Iterable to list.

I noticed and that seems OK for now.

docs/add_new_formats.md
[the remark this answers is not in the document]
lyuyangh: Sounds good! I have renamed them.

PR #230, opened 2022-04-12
Store overall result in dataclass Result instead of a separate field

OscarWang114:
Small modification: 
- Stores the overall analysis result in dataclass `Result`, which is the proper way, instead of `OverallStatistics`.
- Added typing for `Result`


neubig (approved):
LGTM!

PR #232, opened 2022-04-13
Added CI for multiple versions

neubig:
We discussed compatibility of ExplainaBoard with multiple Python versions, and I think we should probably support at least 3.8, 3.9, and 3.10. This PR adds CI for these versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,12 +9,15 @@ on:
 jobs:
   tests:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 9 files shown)


neubig:
This is failing on:
* A syntax error for Python 3.8
* A dependency of a dependency (datalab -> p_tqdm) for Python 3.10

I'll try to fix these. We should decide if we need to support 3.8, but that's actually what I had on my computer so it's probably not old enough to completely rule out.

lyuyangh (approved):
Looks good! Thanks!

PR #244, opened 2022-04-17
F1, Test suits, etc

pfliu-nlp:


neubig:
FYI: "'type' object is not subscriptable" is fixed by adding `from __future__ import annotations` to the top of the file.

pfliu-nlp:
A summary of some major modifications:
* introduce [span class, operation](https://github.com/neulab/ExplainaBoard/blob/generalize-f1-evaluation/explainaboard/utils/span_utils.py), and relaxing evaluation strategy
* more [unit tests](https://github.com/neulab/ExplainaBoard/blob/generalize-f1-evaluation/explainaboard/tests/test_ner.py) for NER task and span-based implementation
* fixed some bugs
* caching [max/min value](https://github.com/neulab/ExplainaBoard/pull/244/commits/313334ef070905f07cd4d642690e517e17d97078) of features for adaptive bucket interval

Some things will be put in further PR:
* deduplicate this [span class](https://github.com/neulab/ExplainaBoard/blob/generalize-f1-evaluation/explainaboard/utils/span_utils.py) and figure out how to achieve efficient bucket-level relaxing evaluation.
* introduce sequence labeling class
* adding unit tests for all tasks
* add docs

explainaboard/metric.py
```diff
@@ -151,7 +152,13 @@ def aggregate_stats(self, stats: MetricStats) -> np.ndarray:
         :param stats: stats for every example
         :return: aggregated stats
         """
-        return np.mean(stats.get_data(), axis=0)
+
+        res = copy.deepcopy(stats)
```
neubig, line 156: We don't need a deepcopy here, do we? Couldn't we just do `data = get_data()` and then use `data` for the rest of the function? I'd like to avoid copying memory unnecessarily.

explainaboard/metric.py
```diff
@@ -151,7 +152,13 @@ def aggregate_stats(self, stats: MetricStats) -> np.ndarray:
...
+
+        if res.get_data().size == 0:
+            return np.array(0.0)
```
neubig, line 159: I'm assuming this is a special case for buckets where the number of examples is zero.

As an alternative suggestion: could we make this `np.zeros(self.stats_length)` where `self.stats_length` is the expected number of columns the sufficient statistics? This would reduce the necessity to make special cases throughout the rest of the code.

explainaboard/metric.py
```diff
@@ -329,12 +337,18 @@ def calc_stats_from_data(
     def calc_metric_from_aggregate(
         self, agg_stats: np.ndarray, config: Optional[MetricConfig] = None
     ) -> float:
+
+        if agg_stats.size == 1:
```
neubig, line 341: If the above change is implemented this line can be removed.

explainaboard/metric.py
```diff
@@ -367,8 +382,8 @@ class BIOF1Score(F1Score):
 
     def calc_stats_from_data(
```
neubig, line 383: Could we merge `BIOF1Score` and `BMEF1Score` into a single class and do the conversion to `list[list[Span]]` in the appropriate places before passing it into the metric function? It seems that this would be quite a bit simpler.

explainaboard/processors/conditional_generation.py
```diff
@@ -1,6 +1,5 @@
 from __future__ import annotations
```
neubig, line 1: Could we revert the modifications to this file and save them for a future PR? I understand what you're doing here, but I'd like to discuss the design of this a bit more, but and don't want to hold up merging this PR. As far as I can tell this wouldn't be too harmful to the other parts.

explainaboard/processors/named_entity_recognition.py
```diff
@@ -351,12 +328,33 @@ def _complete_features(
...
             # Get values of bucketing features
             tokens = dict_sysout["tokens"]
 
             # sentence_length
             dict_sysout["sentence_length"] = len(tokens)
+            self._get_max_min_value(sys_info, "sentence_length", len(tokens))
+
+            # entity density
+            dict_sysout["entity_density"] = len(
+                BIOSpanOps().get_spans(tags=dict_sysout["true_tags"])
```
neubig, line 353: `BIOSpanOps` can probably be initialized outside of the for loop?

explainaboard/processors/named_entity_recognition.py
```diff
@@ -351,12 +328,33 @@ def _complete_features(
...
+        for x in active_features:
+            (sent_feats if (x in sys_features) else tok_feats).append(x)
+
         for _id, dict_sysout in tqdm(enumerate(sys_output), desc="featurizing"):
             # Get values of bucketing features
             tokens = dict_sysout["tokens"]
 
             # sentence_length
             dict_sysout["sentence_length"] = len(tokens)
+            self._get_max_min_value(sys_info, "sentence_length", len(tokens))
```
neubig, line 349: Couldn't we just do this calculation of the min and max when within the bucketing functions in `bucketing.py`? That seems much simpler than having to do it in the for loop each time we do feature extraction for each task.

explainaboard/processors/named_entity_recognition.py
```diff
@@ -415,54 +436,56 @@ def _bucketing_samples(
...
         )
 
         # Bucketing
-        samples_over_bucket_true = {}
         samples_over_bucket_pred = {}
-        performances_over_bucket = {}
-        for feature_name in tqdm(bucket_features, desc="bucketing"):
-
+        for feature_name in tqdm(tok_feats, desc="span-level bucketing"):
             # Choose behavior based on whether this is a feature of samples or spans
```
neubig, line 452: This comment is out of date?

explainaboard/processors/named_entity_recognition.py
```diff
@@ -415,54 +436,56 @@ def _bucketing_samples(
...
-                feature_name, sys_output, lambda x: x["true_entity_info"]
+
+            feat_dict = self._get_feature_dict(
+                sys_output, feature_name, lambda x: x['true_entity_info']
             )
-            feat_dict = {x: y for x, y in zip(span_ids_true, feat_vals)}
+
             samples_over_bucket_true[feature_name] = bucket_func(
                 dict_obj=feat_dict,
+                max_value=my_feature.max_value
```
neubig, line 465: Calculating within the bucketing function would also remove the necessity to do this.

explainaboard/processors/named_entity_recognition.py
```diff
@@ -509,16 +533,26 @@ def get_bucket_cases_ner(
...
-        for pos, tags in sample_dict.items():
+        for span, tags in sample_dict.items():
             true_label = tags.get('true', 'O')
             pred_label = tags.get('pred', 'O')
 
-            split_pos = pos.split("|||")
-            sent_id = int(split_pos[0])
-            span = split_pos[-1]
-            system_output_id = sys_output[int(sent_id)]["id"]
+            # true_span_triple = (
```
neubig, line 540: Remove comment.

explainaboard/processors/named_entity_recognition.py
```diff
@@ -546,20 +580,21 @@ def get_bucket_performance_ner(
         :return: bucket_name_to_performance: a dictionary that maps bucket names to
             bucket performance
         """
-
+        # getattr(explainaboard.metric, f'BIO{name}')
         metric_names = unwrap(sys_info.metric_names)
         config = explainaboard.metric.F1ScoreConfig(ignore_classes=['O'])
         bucket_metrics = [
-            getattr(explainaboard.metric, name)(config=config) for name in metric_names
+            getattr(explainaboard.metric, f'{name}')(config=config)
```
neubig, line 587: Just `name` is fine.

explainaboard/processors/named_entity_recognition.py
```diff
@@ -546,20 +580,21 @@ def get_bucket_performance_ner(
         :return: bucket_name_to_performance: a dictionary that maps bucket names to
             bucket performance
         """
-
+        # getattr(explainaboard.metric, f'BIO{name}')
```
neubig, line 583: Remove comment.

explainaboard/processors/named_entity_recognition.py
```diff
@@ -546,20 +580,21 @@ def get_bucket_performance_ner(
...
         ]
 
         bucket_name_to_performance = {}
         for bucket_interval, spans_true in samples_over_bucket_true.items():
 
             if bucket_interval not in samples_over_bucket_pred.keys():
                 raise ValueError("Predict Label Bucketing Errors")
-            else:
-                spans_pred = samples_over_bucket_pred[bucket_interval]
+            # else:
```
neubig, line 596: Remove comment.

explainaboard/processors/processor.py
```diff
@@ -228,6 +227,37 @@ def _init_customized_features(self, metadata: dict):
                 else:
                     raise NotImplementedError
 
+    def _get_max_min_value(
```
neubig, line 230: This seems to not be necessary, let's just take the min and max in the bucketing functions.

explainaboard/tests/utils/test_BIOSpanOps.py
```diff
@@ -0,0 +1,91 @@
+import unittest
```
neubig, line 1: File names should be lowercase: `test_bio_span_ops.py`

explainaboard/utils/eval_basic.py
```diff
@@ -0,0 +1,132 @@
+from __future__ import annotations
```
neubig, line 1: Do we need this file?

explainaboard/utils/eval_basic_ner.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import annotations
```
neubig, line 1: Do we need this file?

explainaboard/utils/eval_bucket.py
```diff
@@ -0,0 +1,11 @@
+def f1_seqeval_bucket(pred_chunks, true_chunks):
```
neubig, line 1: Do we need this file?

neubig (changes requested):
I did the review! As a meta-comment, somewhat more granular commits would be a bit easier to review, but it's not a big deal.

Overall comments:
1. Let's rename "Chinese Word Segmentation" to "Word Segmentation", as it's also applicable to a few other languages such as Japanese and Thai (morphological segmentation for other languages is a bit different, as word-boundaries are given).
2. The newly introduced files in `explainaboard/tests/artifacts/` are somewhat big. Would a smaller file be sufficient? They're not so big it's a huge problem, but in general tests can be on somewhat minimal files I think.
3. I like the tests in NER. Thanks!

docs/add_new_tasks.md
```diff
@@ -108,6 +108,32 @@ For example, in this file `explainaboard/processors/__init__.py`, we have:
 from . import text_classification
 ```
 
+
+## Update new task information
+Once you finish all the above steps, you need to register the new task in the script tasks.py by declaring the supported formats, metrics, etc.
+
+For example, the following registered information is made for text classification tasks.
```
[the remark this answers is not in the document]
pfliu-nlp: thanks and done!

docs/final_check_for_contribution.md
```diff
@@ -0,0 +1,20 @@
+# Final Check when You're Contributing as SDK Developers
+
+
+Hi, first, thanks for your brilliant contribution to ExplainaBoard SDK :smiley: !
+
+This doc provides some friendly reminder for ExplainaBoard developers so that all relevant scripts and docs
```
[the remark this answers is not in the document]
pfliu-nlp: thanks and done!

explainaboard/feature.py
```diff
@@ -307,6 +307,8 @@ class ClassLabel:
     names: Optional[list[str]] = None
     description: Optional[str] = None
     names_file: Optional[str] = None
+    max_value: Optional[int] = None
```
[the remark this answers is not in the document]
pfliu-nlp: So far, we calculate them when performing `_complete_features`, and then save them in the Feature class. This should be the most effective way. 
>"or can we just take the "max" and "min" dynamically?"

Storing them in Feature class has two benefits: (1) efficiency (2) easier for different functions to use

explainaboard/loaders/kg_link_tail_prediction.py
```diff
@@ -56,13 +59,26 @@ def load(self) -> list[dict]:
             self._output_data, self._output_source
         )
 
+        # Map entity into an interpretable version
+        scriptpath = os.path.dirname(__file__)
+        entity_dic = {}
+        with open(
+            os.path.join(scriptpath, '../pre_computed/kg/entity2wikidata.json'),
```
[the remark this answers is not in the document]
pfliu-nlp: Yes we can, maybe we could cache [this file](https://github.com/neulab/ExplainaBoard/blob/main/explainaboard/pre_computed/kg/entity_type_level_map.json) as well, and remove the pre_computed folder.  (new issue: https://github.com/neulab/ExplainaBoard/issues/251#issue-1211558025)

explainaboard/metric.py
pfliu-nlp: no probolem and done

explainaboard/processors/named_entity_recognition.py
pfliu-nlp: I'm relatively sure it would be significantly slower especially when (1) we have multiple features (2) large test samples and (3) bucketing is performed on-the-fly.   This is based on some lessons from DataLab's bucketing experience.

explainaboard/processors/named_entity_recognition.py
pfliu-nlp: good capture.

explainaboard/processors/processor.py
pfliu-nlp: Similar to the above reply.  I tend to keep it.

explainaboard/utils/eval_basic_ner.py
pfliu-nlp: tested: we don't need it, let's delete it.

explainaboard/utils/eval_basic.py
pfliu-nlp: tested: we don't need it, let's delete it.

explainaboard/utils/eval_bucket.py
pfliu-nlp: deleted

explainaboard/processors/named_entity_recognition.py
pfliu-nlp: good capture

explainaboard/metric.py
pfliu-nlp: Yes, this is a good suggestion. I actually plan to do it when I introduce the concept of "sequence labeling" task processor, which could make some code mergeable, for example:
* processors of `named entity recognition and `word segmentation` 
* evaluation metrics like `BIOF1Score` `BMESF1Score`

explainaboard/processors/conditional_generation.py
pfliu-nlp: OK, I can do it and done.

explainaboard/metric.py
pfliu-nlp: aha, I'm not sure if I completely get your point so I leave this un-implemented first.

explainaboard/feature.py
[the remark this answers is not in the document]
neubig: I see. But calculating a max is not a very slow operation, right? So it seems that (1) is not a very serious concern.

On the other hand, caching the min and max adds a significant amount of complexity to the code. If we did it in the bucketing functions it would add maybe 2 lines of code in each bucketing function, right? Instead, we've added about 50-100 lines of code in this PR just for calculating the max.

pfliu-nlp:
>"The newly introduced files in explainaboard/tests/artifacts/ are somewhat big. Would a smaller file be sufficient?"

Yes, I do it intendedly since I want to test training-set dependent features. But of course, for some other tasks, I plan to do it locally (but keep one alive in our unit test should be necessary)

neubig (approved):
OK, I think this is OK. We can fix the min/max stuff later I guess.

PR #248, opened 2022-04-20
Optimizes bootstrap sampling

neubig:
Some profiling showed that bootstrap sampling was a bottleneck in the speed of bucketing.

This small fix greatly improves the speed of bucketing:
Before: 1.6s on my macbook
After: 0.62s on my macbook


OscarWang114 (approved):
This is awesome, thanks! Looking forward to the speed up in web interface.

PR #249, opened 2022-04-20
add split to SysOutputInfo and auto generate file types for each task

lyuyangh:
1. Added `dataset_split` to `SysOutputInfo`. We need to store this information of each system for the leaderboard feature. 
    - The metadata dict for `get_processor` is getting increasingly complicated. I think it might be helpful to define a dataclass for it. It seems that metadata basically supports a subset of `SysOutputInfo` fields. Maybe `SysOutputInfo` can be a derived class of `Metadata`? Or maybe we can make metadata an attribute of `SysOutputInfo` so it's clear which part is metadata (provided by the user) and which part is analysis report (generated by explainaboard)?
2. replaced hard coded file types with a function that extracts supported file types from the loader definitions
    - I had to move `TaskType` to constants.py to avoid circular dependency
3. replaced a validation exception with ValueError so I can properly handle it in the web app.

This PR also releases a new version (0.8.9)


explainaboard/convert.py
```diff
@@ -0,0 +1,42 @@
+import csv
```
neubig, line 1: This seems like a script that shouldn't be checked in?

neubig (approved):
Looks great, but I think there's one script you didn't intend to check in perhaps?

explainaboard/convert.py
lyuyangh: Thanks! And yeah, I didn't notice I checked in that script too. I have removed it in a new commit.

Issue #251, opened 2022-04-21
Download files in `pre_computed` folder on-the-fly and cache them


pfliu-nlp:
As title goes and some discussions: https://github.com/neulab/ExplainaBoard/pull/244#discussion_r855298802

PR #252, opened 2022-04-21
Remove pre-computed directory and cache

neubig:
Fixes https://github.com/neulab/ExplainaBoard/issues/251


pfliu-nlp:
looks nice

PR #254, opened 2022-04-22
Vectorize bucketing

neubig:
I took a stab at converting bucketing to use numpy vectorization and calculates min/max within the bucketing function. Here are the metrics on NER processing:

Featurization time: 2.0iter/s -> 3.0iter/s (50% faster)
Bucketing time: 3.95s/iter -> 3.97s/iter (a little slower)
Code: simpler

in addition, I reduced the number of samples for bootstrap sampling from 2000 by default to 1000 by default. This changes
Bucketing time: 3.97s/iter -> 2.04s/iter


pfliu-nlp (approved):
looks nice!

PR #258, opened 2022-04-23
Fix bug feature type for multiple qa

pfliu-nlp:
this should be useful for fixing this bug: https://github.com/neulab/explainaboard_web/issues/127#issuecomment-1105598033


pfliu-nlp:
>"https://github.com/neulab/ExplainaBoard/runs/6137534624?check_suite_focus=true"

I actually didn't get the reason, I guess it is because eaas is busy at that time.

neubig (approved):
Very simple, I approve!

PR #260, opened 2022-04-23
fix: duplicate task category name

lyuyangh:
Related to the "strange behavior of task drop-down box when submitting a system output" bug reported in https://github.com/neulab/explainaboard_web/issues/127
A fix has already been provided in https://github.com/neulab/explainaboard_web/pull/135 (the frontend) so we don't necessarily have to enforce task category names to be unique in the SDK but I think this may be a convenient assumption for the tasks list?


pfliu-nlp:
Thanks for debugging this!

pfliu-nlp (approved):

Issue #239, opened 2022-04-15
Better support for specifying dataset splits from DataLab


neubig:
Currently ExplainaBoard basically assumes that we're analyzing the "test" set from DataLab.

However, we may want to analyze other dataset splits, and some major datasets (e.g. SQuAD) don't even release the test set. We should support specifying the dataset split programmatically, including through the CLI.

PR #265, opened 2022-04-25
Fix metaphor QA

neubig:
This re-adds metaphor QA. It also adds the ability to specify splits from DataLab, fixing https://github.com/neulab/ExplainaBoard/issues/239


OscarWang114 (approved):
Looks good!

PR #266, opened 2022-04-25
Fixes metric configs

neubig:
This makes two changes:

1. Small change that all metric config classes now keep track of their class name, so it's written out into a report. This is good because you can then easily recreate the metric config by doing the following.

    getattr(explainaboard.metric, metric_config_dict['cls'])(**metric_config_dict)

2. Larger change that we now removed most of the "supported" things from `tasks.py`. The new canonical way of getting the supported processors and loaders is: `explainaboard.loaders.loader_registry.get_supported_file_types_for_loader(TaskType)` and `explainaboard.processors.processor_registry.get_metric_list_for_processor(TaskType)`


OscarWang114:
@neubig I'm trying to execute your first bullet point:
```
mt = explainaboard.tasks.get_task_categories()[0].tasks[0].name
>>> type(mt)
<TaskType.machine_translation: 'machine-translation'>
>>> cf = explainaboard.processors.processor_registry.get_metric_list_for_processor(mt)[0]
>>> type(cf) 
<class 'explainaboard.metric.EaaSMetricConfig'>
>>> getattr(explainaboard.metric, cf['cls'])(**cf)
  File "<stdin>", line 1, in <module>
KeyError: 'cls'
```

I believe `EaaSMetricConfig` inherits `dict`, so I'm not sure why this is now working.

OscarWang114 (approved):
LGTM, this is awesome. Two very useful features for the web interface. 👍

PR #271, opened 2022-04-27
Added language modeling task

neubig:
This adds language modeling as a task that can be analyzed with ExplainaBoard.


explainaboard/metric.py
```diff
@@ -781,6 +781,54 @@ def sample_level_metric(
         return f1
 
 
+@dataclass
```
pfliu-nlp, line 784: We probably need a doc to show how to add a new metric.

pfliu-nlp (approved):

PR #275, opened 2022-04-28
add some docs about how to use ExplainaBoard Web

pfliu-nlp:
I will add some docs about how to use ExplainaBoard Web for our collaborators from the NEC lab.
(We can move  such docs to some other places later if needed, but now I think this (ExplainaBoard SDK repo) would be a good place)


neubig (approved):
Looks great, I just did a little bit of proofreading.

PR #277, opened 2022-04-29
Add support for code generation

neubig:
Added support for analyzing code generation, with CoNaLa as an example.


pfliu-nlp (approved):
looks good

PR #278, opened 2022-04-29
Enable calculating of MR and MRR via user-provided rank information

hwidjaja:


explainaboard/loaders/kg_link_tail_prediction.py
```diff
@@ -112,6 +115,9 @@ def load(self) -> list[dict]:
                             "gold_" + features_dict["predict"]
                         ],  # the entity to which we compare the predictions
                         "predictions": features_dict["predictions"],
+                        "true_rank": features_dict["true_rank"]
+                        if "true_rank" in features_dict.keys()
+                        else None,  # rank of the true entity in predictions
```
neubig, line 120:
```suggestion
                        "true_rank": features_dict.get("true_rank", None),
```

explainaboard/processors/kg_link_tail_prediction.py
```diff
@@ -177,6 +181,34 @@ def _statistics_func(self, samples: Iterator[dict[str, str]]):
...
+
+        metrics = unwrap(self._get_metrics(sys_info))
+        true_data = [self._get_true_label(x) for x in sys_output]
+        pred_data = [self._get_predicted_label(x) for x in sys_output]
+        rank_data = [
+            self._get_rank_data(x) for x in sys_output
+        ]  # rank of true entity in predictions
+        assert all(
+            item is not None for item in rank_data
+        ), 'Some data points do not have rank information; check your system outptus.'
```
neubig, line 202:
```suggestion
        if any(item is None for item in rank_data):
            raise ValueError('Some data points do not have rank information; check your system outputs.')
```

explainaboard/processors/kg_link_tail_prediction.py
```diff
@@ -177,6 +181,34 @@ def _statistics_func(self, samples: Iterator[dict[str, str]]):
...
+
+        metric_stats = []
+        for metric in metrics:
+            if isinstance(metric, MeanReciprocalRank) or isinstance(metric, MeanRank):
+                metric_stats.append(metric.calc_stats_from_rank(rank_data))
+            else:
+                metric_stats.append(metric.calc_stats_from_data(true_data, pred_data))
```
neubig, line 209: Hits can also be calculated from rank right? Probably all of them should be calculated from the rank to be consistent and simplify the code?

explainaboard/processors/kg_link_tail_prediction.py
```diff
@@ -283,3 +315,6 @@ def _get_true_label(self, data_point: dict):
 
     def _get_predicted_label(self, data_point: dict):
         return data_point["predictions"]
+
+    def _get_rank_data(self, data_point: dict):
+        return data_point.get("true_rank", None)
```
neubig, line 320: Suggestion: if `true_rank` is not included in the output, could we try to get it from the system predictions before returning None? This would make it more flexible to be able to handle when ranks are not passed in, such as when the predictions always contain the true answer.

One issue with requiring `true_rank` to be provided is that it prevents us from being able to use blind test sets.

docs/task_kg_link_tail_prediction.md
```diff
@@ -173,22 +179,21 @@ pythonic interface provided by ExplainaBoard.
 ### Customized Bucket Order
 In some situation, users aim to specify the bucket order according to their needs. Following [code](https://github.com/neulab/ExplainaBoard/blob/8ccd1a71531bc3b9e2f9e539cb001353cc49ebca/explainaboard/tests/test_kg_link_tail_prediction.py#L60) gives an example.
 
-
+Below is an example of sorting alphabetically by bucket value:
```
neubig, line 182: Overall comment: these code examples seem a bit verbose for a `.md` file. If they could be made a bit more concise that'd be nice, but it's not an urgent issue.

neubig (commented):
Looks great, thanks for the improvements. I've left a few comments.

hwidjaja:
Thanks for the comments Graham!

explainaboard/processors/kg_link_tail_prediction.py
hwidjaja: In the case of blind test sets, would it then be reasonable to assume that the system outputs will contain the entire list of entities as predictions (instead of just the top-k)? Otherwise, there would be no way to calculate the correct MR and MRR. (Hits@k would still be accurate as long as the predictions contain at least k elements.) Thoughts?

neubig: Yes, for now we could throw an error if neither of the two conditions are satisfied. In the future (or even in this PR), we might also add the ability to specify a "default_rank" for all things that were not included. Then the condition would be:

```
if true rank exists:
  return true rank
elif true_label in n-best list:
  return rank of true_label in n-best list
elif default_rank is not None:
  return default_rank
else:
  throw error
```

neubig (approved):
I approve if you think this is ready!

PR #280, opened 2022-05-01
Make CoNLL loader work with space-separated files

neubig:
The CoNLL loader wasn't working with space-separated files for predictions, so this PR fixes and tests this.

### DIFF
--- a/explainaboard/loaders/file_loader.py
+++ b/explainaboard/loaders/file_loader.py
@@ -236,22 +236,26 @@ class CoNLLFileLoader(FileLoader):
                     field.src_name: [] for field in self._fields
                 }  # reset
 
+        max_field: int = max([narrow(x.src_name, int) for x in self._fields])
         for line in raw_data:
             # at sentence boundary
             if line.startswith("-DOCSTART-") or line == "" or line == "\n":
                 add_sample()
             else:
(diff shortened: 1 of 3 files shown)


explainaboard/loaders/file_loader.py
```diff
@@ -236,22 +236,26 @@ def add_sample():
...
-                            f"not enough fields for {line} (sentence index: {guid})"
-                        )
+                if len(splits) <= max_field:  # not separated by tabs or spaces
+                    raise ValueError(
+                        f"not enough fields for {line} (sentence index: {guid})"
+                    )
 
                 for field in self._fields:
+                    fid = narrow(field.src_name, int)
+                    if fid >= len(splits):
```
lyuyangh, line 255: Thanks for the PR! It looks great! 

(Just one small thing) I'm trying to understand this check here. I think this may not be necessary since `len(splits) > max_field` is checked in the previous if/else block?

lyuyangh (approved):

explainaboard/loaders/file_loader.py
neubig: Thanks, this was indeed not necessary given the checks above. It was something I had introduced in the process of debugging.

PR #281, opened 2022-05-02
Add caching lifetime

neubig:
This commit adds a lifetime for cached files so they will be automatically updated when they exceed a certain age. This will be useful when there are online files we'd like to access and cache, but update occasionaly.

It also bumps the version, as I'd like to use this in ExplainaBoard web.


lyuyangh (approved):
Looks good!

This implementation lets the user decide when to update cached files. I looked at the response of [this](http://phontron.com/download/explainaboard/pre_computed/kg/entity_type_level_map.json) file. I think the server already specifies [Etag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) and [Expires](https://www.geeksforgeeks.org/http-headers-expires/). If we want to have more control over when the file gets updated, we can maybe get the "lifetime" information from the response?

neubig:
Ah, good point! It'd probably be good to get that in addition to letting the user specify.

PR #285, opened 2022-05-03
De-duplicate NER and CWS  & Introduce Sequence Labeling

pfliu-nlp:


explainaboard/processors/sequence_labeling.py
```diff
@@ -0,0 +1,573 @@
+from __future__ import annotations
```
neubig, line 1: There are lots of places in this file that refer to "entity" (e.g. in the feature descriptions). This should probably be generalized to "span".

neubig (approved):
Looks great! Just one small documentation comment.

Also, one thing you might want to consider: I think that what you're calling "sequence labeling" here is actually something more like "span prediction"? This is in contrast to a more traditional sequence labeling task like POS tagging, which assigns one tag to each word. It might be worth renaming sequence labeling to "span prediction" (or maybe not, I'm not sure).

explainaboard/processors/sequence_labeling.py
pfliu-nlp: Thanks! Will clear out all `entity`s. 

>"I think that what you're calling "sequence labeling" here is actually something more like "span prediction"?"
Good point. I also hesitate among `tagging`, `sequence labeling`, `span prediction`.  Let me mark it as an issue.

PR #288, opened 2022-05-04
Revisit/Add customized metadata 

pfliu-nlp:


pfliu-nlp:
BTW, it seems that we have already supported the customized feature's functionality for (multiple_choice_qa](https://github.com/neulab/ExplainaBoard/blob/c3af2de7e06d7147b184efafa7b68c68d29f3b9a/explainaboard/tests/test_qa_multiple_choice.py#L66)

neubig (approved):
LGTM!

PR #295, opened 2022-05-07
Catch missing training datasets

neubig:
Datalab import crashed on datasets that only have a test set. This PR fixes this and throws a warning.


OscarWang114 (approved):
LGTM, and it seems uncontroversial!

PR #298, opened 2022-05-09
rename cls to avoid name conflict

lyuyangh:
This PR renames two places where we used "cls" as attribute names.

When I use openapi codegen to generate a python client for our open APIs, it generates a series of classes. One of them is used to store data in `SysOutputInfo`. And one of the classmethods takes all the attributes as inputs so there is a name conflict. It's difficult to foresee this issue. I think maybe we should avoid python reserved keywords in general (especially for function arguments and attributes).

I think both of these methods are only used in explainaboard_web to serialize/deserialize objects so this shouldn't affect other parts of this codebase. I'll update explainaboard_web and the DB once this PR is merged.

There are two tests that are failing. It looks like both of them are related to the data downloaded from datalab.


pfliu-nlp (approved):
Looks good

PR #299, opened 2022-05-09
Refactor for translation

neubig:
This PR makes a major refactoring to be able to support machine translation (and multilinguality in general) in a good way. It also pays back a lot of technical debt that was accrued. Unfortunately this will cause several breaking changes regarding the interface and file formats, but I think these changes are for the better in general. I raised the minor version number to 0.9.X to reflect that these changes are large.

* Adds support for separate source and target languages including in tokenization and features
* Refactors the JSON input format, renaming `user_defined_metadata_configs` into `metadata` and moving `user_defined_features_configs` into `custom_features` within this `metadata`
* Unify the JSON format for datasets and system inputs, add metadata support to both
* Tries to infer the languages from metadata of the dataset or inputs, defaults to "None" otherwise (instead of assuming English)
* Adds dataloaders that support translation datasets in Datalab (this was a bit complicated because names of fields are the names of languages, dynamic for every dataset)
* Makes the file format of KG Link Prediction follow the same standard as all the other tasks

I'll follow up with a PR fixing explainaboard_web too later today or later this week.


explainaboard/loaders/file_loader.py
```diff
@@ -128,47 +216,108 @@ def generate_id(self, parsed_data_point: dict, sample_idx: int):
...
+                    src_name = field.src_name
+                else:
+                    src_name = [field_mapping.get(x, x) for x in field.src_name]
+                new_field = copy.copy(field)
+                new_field.src_name = src_name
+                fields.append(new_field)
+            return fields
+
+    @classmethod
+    def find_field(cls, data_point, field, field_mapping=None):
```
pfliu-nlp, line 265: It would be nice if a short docstring could be added here.

explainaboard/loaders/file_loader.py
```diff
@@ -306,12 +484,41 @@ def replace_labels(cls, features: dict, example: dict) -> dict:
...
         config = narrow(data, DatalabLoaderOption)
         dataset = load_dataset(
             config.dataset, config.subdataset, split=config.split, streaming=False
         )
-        return [self.replace_labels(dataset.info.features, x) for x in dataset]
+        # TODO(gneubig): patch for an inconsistency in datalab, where DatasetDict
+        #  doesn't have info
+        if isinstance(dataset, DatasetDict) or isinstance(dataset, IterableDatasetDict):
+            raise ValueError('Cannot handle DatasetDict returns')
+        info = dataset.info
```
pfliu-nlp, line 515: Have we added `info` as a member variable into the `DatasetDict` class in DataLab (we have `DatasetDict[split]._info`)? Sometime it should be split-dependent. For example, features of test samples are not necessarily the same as features of training samples.

explainaboard/loaders/file_loader.py
```diff
@@ -306,12 +484,41 @@ def replace_labels(cls, features: dict, example: dict) -> dict:
...
+        # Infer metadata from the dataset
+        metadata = FileLoaderMetadata()
+        if info.languages is not None:
+            metadata.supported_languages = info.languages
+            if (
+                metadata.supported_languages
+                and len(metadata.supported_languages) > 0
```
pfliu-nlp, line 522: (1) it would be nice if docstring is added here
(2) what if `len(metadata.supported_languages)` > 3?

explainaboard/loaders/kg_link_tail_prediction.py
```diff
@@ -68,56 +70,35 @@ def load(self) -> list[dict]:
...
-                            "gold_" + features_dict["predict"]
-                        ],  # the entity to which we compare the predictions
-                        "predictions": features_dict["predictions"],
-                        "true_rank": features_dict.get(
-                            "true_rank", None
-                        ),  # rank of the true entity in predictions
-                    }
-                )
-        return data
+            data.append(data_i)
```
pfliu-nlp: I remembered the above code should have been updated in this PR: https://github.com/neulab/ExplainaBoard/blob/5c423ddd3e21fb14e23cfc9122c832e7732923da/explainaboard/loaders/kg_link_tail_prediction.py#L65
It's a little strange to see the outdated code has been updated again.

data/system_outputs/fb15k-237/no-user-defined.json
```diff
@@ -0,0 +1,142 @@
+[
```
pfliu-nlp, line 1: it seems that a similar file should be added in this PR: https://github.com/neulab/ExplainaBoard/tree/loaders_support_preprocessor/data/system_outputs/fb15k-237

pfliu-nlp: got the answer from the later commits.

explainaboard/loaders/kg_link_tail_prediction.py
pfliu-nlp: got the answer from the later commits.

explainaboard/explainaboard_main.py
```diff
@@ -40,22 +48,21 @@ def get_tasks(task: str, system_outputs: list[str]) -> list[str]:
...
         for sys_output in system_outputs:
-            loader = get_custom_dataset_loader(
-                dummy_task,
-                sys_output,
-                sys_output,
-                dataset_file_type=file_type,
-                output_file_type=file_type,
-            )
-            if not loader.user_defined_metadata_configs:
+            task_or_die: TaskType | None = None
```
pfliu-nlp, line 52: The name of this variable (`task_or_die` ) is so interesting that If want to know how you are going to add a docstring for it :)

explainaboard/explainaboard_main.py
```diff
@@ -16,21 +15,30 @@
...
+def get_tasks(task: TaskType, system_outputs: list[str]) -> list[TaskType]:
+    """
+    Get the task for each system output.
+    :param task: Explicitly specified task. Use if present
+    :param system_outputs: System output files, load from metadata in these files if
+      an explicit task is not set
+    :return: A list of task types for each system
     """
-    real_tasks: list[str]
+    real_tasks: list[TaskType] = []
```
pfliu-nlp, line 41: We all realize that it's probably a little stupid to get tasks (using a dummy task) in this way, maybe in the future we could think about some elegant way to solve this issue fundamentally.

explainaboard/loaders/file_loader.py
neubig: I don't think this exists as a member variable in DatasetDict (it threw a `mypy` error). It probably should be fixed in DataLab.

explainaboard/loaders/loader.py
```diff
@@ -110,72 +104,21 @@ def select_file_loader(
...
         )
+        dataset_loaded_data.metadata.merge(output_loaded_data.metadata)
         if len(dataset_loaded_data) != len(output_loaded_data):
             raise ValueError(
                 "dataset and output are of different length"
                 + f"({len(dataset_loaded_data)} != {len(output_loaded_data)})"
             )
-        for i, output in enumerate(output_loaded_data):
+        data_list: list[dict] = output_loaded_data.raw_data
+        for i, output in enumerate(data_list):
```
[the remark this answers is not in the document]
neubig: The customized features are saved in `dataset_loaded_data.metadata` for use later.

explainaboard/explainaboard_main.py
neubig: Your wish is my command (in the next commit ;)).

explainaboard/explainaboard_main.py
neubig: I think that I actually solved it by just reading in the JSON without using a loader?

explainaboard/loaders/file_loader.py
pfliu-nlp: yeah, by design, there is no definition of `info` for DatasetDict but there is for `Dataset`. But we can introduce `info` for DatasetDict as well,  just highlighting a caveat as above: `Sometime it should be split-dependent. For example, features of test samples are not necessarily the same as features of training samples.`

explainaboard/explainaboard_main.py
pfliu-nlp: >"Or did you mean you wanted to solve it without reading the JSON twice?"

yes, maybe it's impossible if we don't introduce any more arguments.

PR #303, opened 2022-05-12
Rename some task names

pfliu-nlp:


explainaboard/tasks.py
```diff
@@ -130,7 +134,7 @@ class TaskCategory:
         ],
     ),
     TaskCategory(
-        "kg-link-tail-prediction",
+        "kg_prediction",
```
neubig, line 137: I know this is WIP, but I was looking at it out of curiosity and noticed that this name uses underbars while other names use hyphens. This was also the case in DataLab as well. Maybe we can consolidate to hyphens?

pfliu-nlp: Yes, this is a typo, and thanks for capturing it!

pfliu-nlp:
Sister PR: https://github.com/ExpressAI/DataLab/pull/199

neubig (approved):
Looks good, thanks!

Issue #306, opened 2022-05-16
Unify naming of "model" and "system"?


neubig:
Currently we're mixing the use of "model" and "system". "system" is more common, but "model" is also in various places, such as `explainaboard_main`: https://github.com/neulab/ExplainaBoard/blob/main/explainaboard/explainaboard_main.py#L147

I think "system" is clearly more accurate, as it's the whole system that ends up producing an output, not just the model. So maybe we should unify to "system"? It would just be renaming, but unfortunately this touches many places in the code so we'd need to make sure we don't break things. I've actually been thinking about this for a little while now but was busy with other things and didn't have time to mention it.

@pfliu-nlp , what do you think?

pfliu-nlp:
cool, have a similar feeling, I think we can go with `system`!

PR #307, opened 2022-05-16
Renames "model" to "system"

neubig:
Fixes https://github.com/neulab/ExplainaBoard/issues/306

There will be follow-up PRs to make explainaboard_web and explainaboard_client compliant.


pfliu-nlp (approved):
looks good to me

PR #314, opened 2022-05-19
Fixed features for condgen

neubig:
Fixes custom features for conditional generation. Draft because it still needs tests.

### DIFF
--- a/explainaboard/processors/conditional_generation.py
+++ b/explainaboard/processors/conditional_generation.py
@@ -388,8 +388,14 @@ class ConditionalGenerationProcessor(Processor):
         )
         for bucket_feature in active_features:
             if bucket_feature in sys_features:
+                feature_info = sys_features[bucket_feature]
+                feature_func = (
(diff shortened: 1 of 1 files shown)


explainaboard/tests/test_machine_translation.py
```diff
@@ -75,6 +76,31 @@ def test_default_features_dont_modify_condgen(self):
         # condgen features are a subset of MT features
         self.assertDictEqual(mt_features, {**mt_features, **condgen_features_1})
 
+    def test_custom_features(self):
+        loader = get_custom_dataset_loader(
```
pfliu-nlp, line 80: Actually, this type of test for `customized features` is relatively shallow (That's part of the reason why we didn't capture the bug of not using `customized` features). A better idea is to see if this feature has been put into use in the final analysis report. But this could be done in future PR.

pfliu-nlp (approved):
Overall, looks nice.

neubig:
Oh gotcha, it was easy enough to add a test so I did it now, thanks. I also bumped the version and will merge when tests pass.

PR #316, opened 2022-05-22
Refactoring bucket cases and bucketing

neubig:
This PR started out as an attempt to make it possible to record the required information to visualize bucket cases for span- or token-based tasks. However, it ended up being a more major refactoring of bucket cases and bucketing in general, which hopefully nonetheless is positive. Here are the major changes:

* Restructure buckets from `dict[str, BucketPerformance]` to just `list[BucketPerformance]`, as the name of the bucket is included in `BucketPerformance` already.
* For better global consistency, change all string bucket names to tuple bucket intervals.
* Much more extensive type annotation regarding bucketing.
* Refactor BucketCase objects so they don't use strings but actual data structure corresponding to the semantics of the variable (e.g. int if the variable is an int).
* Change processors so they first generate BucketCase objects, then bucket the BucketCase objects (everywhere)
* Mostly merge `BIOSpan{Ops,Metric,MetricConfig}` and `BMESSpan{Ops,Metric,MetricConfig}`


pfliu-nlp (commented):
It would be nice if you could generate a new analysis report (upload them [here](https://github.com/neulab/ExplainaBoard/tree/bucket_case_spans/data/reports) for example), which will help me better understand what happened.

explainaboard/tests/test_ner.py
```diff
@@ -95,59 +95,37 @@ def test_datalab_loader(self):
...
-            second_interval[0],
+            second_bucket.bucket_interval[0],
             0.007462686567164179,
             4,
             "almost equal",
         )
         self.assertAlmostEqual(
-            second_interval[1],
-            0.8603351955307262,
+            second_bucket.bucket_interval[1],
```
pfliu-nlp, line 107: it seems that the number value is a little different from the one in the previous version

explainaboard/utils/span_utils.py
```diff
@@ -156,318 +225,113 @@ def get_matched_spans(
...
-        for i, tok in enumerate(tags_ngrams):
-            tags_table[tok[0]].append(i)
-
-        span_dics = get_span_from_ngrams(tags_ngrams, tags_other_table, len(tags))
-        span_dics_other = get_span_from_ngrams(
-            tags_other_ngrams, tags_table, len(tags_other)
-        )
-
-        return span_dics, span_dics_other
+# TODO(gneubig): this is not used anywhere, so decide to keep or delete
```
pfliu-nlp, line 228: Aha, this is originally expected to be applied to conditional text generation tasks, extending token-level accuracy to span-level

explainaboard/utils/span_utils.py
```diff
@@ -156,318 +225,113 @@ def get_matched_spans(
...
-                    else span.span_text
-                )
-                span.span_econ = self.resources["econ_dic"].get(
-                    f'{lower_text}|||{lower_tag}', 0.0
-                )
-                span.span_efre = self.resources["efre_dic"].get(lower_text, 0.0)
-            spans.append(span)
-
-        return spans
+    def _span_ends(self, tags: list[str], i: int) -> bool:
```
pfliu-nlp, line 330: the introduced strategy of using `_span_ends` and `_span_start` is surprisingly fantastic

pfliu-nlp: Learned: `if one does not find a unified method, it is just because the disassembly is not detailed enough`

pfliu-nlp (approved):
Thanks for a lot of stuff. I finally  finished the reviewing (learning)

explainaboard/tests/test_ner.py
neubig: Yes, the reason for this is because previously bucketing was performed separately on true and predicted spans, but now it is performed on the union of both. I'm pretty sure it's not a bug, but the buckets will be a little bit different.

explainaboard/utils/span_utils.py
neubig: I'll leave it commented out for now, but we can either delete or resurrect it later.

PR #324, opened 2022-06-01
Text regression

qinyiwei:
Add text regression task. Implement three metrics: SegKtauCorr, SysPearsonCorr, RootMeanSquareError.


pfliu-nlp:
Hey, @qinyiwei  thanks a lot for the PR. It would be nice if you could fix these continuous integration errors based on corresponding suggestions.

neubig:
To make it easier to fix those errors, I'd suggest running `pre-commit install`, which will install pre-commit checks for formatting and type-checking. I think these things should be easy to check.

pfliu-nlp:
These conflicts could be resolved manually.

explainaboard/loaders/text_regression.py
```diff
@@ -0,0 +1,70 @@
...
+                    FileLoaderField(5, target_field_names[5], str),
+                    FileLoaderField(6, target_field_names[6], str),
+                    FileLoaderField(7, target_field_names[7], str),
+                ],
+            ),
+        }
+
+    @classmethod
+    def default_output_file_loaders(cls) -> dict[FileType, FileLoader]:
+        field_name = [
```
pfliu-nlp, line 53: To make the naming format consistent with others, we can lowercase these names and use "_" to connect two tokens if needed. For example `TestSet` -> `test_set`

explainaboard/tasks.py
```diff
@@ -157,6 +157,16 @@ class TaskCategory:
             )
         ],
     ),
+    TaskCategory(
+        "text-regression",
+        "evaluate evaluation metrics.  ",
+        [
+            Task(
+                name=TaskType.text_regression,
```
pfliu-nlp, line 197: we could re-consider the task name, some other options, such as `meta-evaluation`

docs/task_text_regression.md
```diff
@@ -0,0 +1,37 @@
...
+
+We have an example dataset file:
+* [data.tsv](./data/system_outputs/text_regression/wmt20-DA/cs-en/data.tsv)
+
+More dataset files can be found at [WMT-DA-20](https://drive.google.com/drive/u/0/folders/1JXpo0yxPLYlNgLbOfP1bzs9z6SOx76Wo).
+
+In order to perform analysis of your results, your system outputs should add one column of system score at the base of dataset file:
+
+```
+SYSName \t SEGID \t TestSet \t src \t ref \t sys \t manualRaw \t manualZ \t systemScore
```
pfliu-nlp, line 19: maybe we can explain a little bit about each column

pfliu-nlp:
hi, @qinyiwei
>"explainaboard   --task text-regression --custom_dataset_paths ./data/system_outputs/text_regression/wmt20-DA/cs-en/data.tsv --system_outputs ./data/system_outputs/text_regression/wmt20-DA/cs-en/score.tsv --output_file_type tsv --output_dir output/cs-en --source_language en --target_language en --metrics SysPearsonCorr SegKtauCorr RootMeanSquareError"

This seems cannot work, detailed logs:

d5e73b377fe653710aeb05ae10826b0

907bedb3fffb593b5ff8ca8a1d0befd

neubig (changes requested):
Thanks a lot for doing this!

My biggest comment is that I think we should rename the task as `nlg-meta-evaluation`, as @pfliu-nlp mentioned. When I see `text-regression`, I would expect just one textual input and one continuous-valued output. For example, we could implement this with `sst5` 5-level sentiment analysis (in a future PR).

Once the name is changed I'll do a more complete review!

pfliu-nlp:
Hi, @neubig just re-named it.

explainaboard/metrics/regression.py
```diff
@@ -0,0 +1,359 @@
...
+import numpy as np
+from scipy.stats import pearsonr
+
+from explainaboard.metrics.metric import Metric, MetricConfig, MetricStats
+from explainaboard.metrics.registry import register_metric_config
+
+
+@dataclass
+@register_metric_config
+class SegKtauCorrConfig(MetricConfig):
```
neubig, line 15: (1) Could `no_human` and `threshold` be documented?
(2) Maybe we should consider the default for `threshold` being `0`? `25` seems a bit arbitrary and opaque.
(3) It'd be nicer if we could avoid having `Seg` be part of the class name, and rather define `corr_level = 'seg'` or `corr_level = 'sys'` as a variable in the configuration. That would make it easier to calculate `Ktau`, `Pearson`, `Spearman` etc. with respect to either segment-level or system-level correlation.

explainaboard/metrics/regression.py
```diff
@@ -0,0 +1,359 @@
...
+                        conc += 1
+                    else:
+                        disc += 1
+                    num += 1
+                else:  # system i is worse than system j
+                    if score[i][1] < score[j][1]:
+                        conc += 1
+                    else:
+                        disc += 1
+                    num += 1
```
neubig, line 112: The following suggestion is just to make things a bit more concise:
```suggestion
                if (abs(score[i][0] - score[j][0]) >= config.threshold):
                    manual_better = score[i][0] > score[j][0]
                    system_better = score[i][1] > score[j][1]
                    if manual_better == system_better:
                        conc += 1
                    else:
                        disc += 1
                    num += 1
```

neubig (commented):
Thanks for the quick action, and overall looking pretty good! I have a few comments.

First, we can probably move the directory:
`data/system_outputs/text_regression`
->
`data/system_outputs/nlg_meta_evaluation`

explainaboard/metrics/regression.py
pfliu-nlp: fixed: `Regarding (1) (2) will fix them once I get feedbacks from Yiwei`

neubig (approved):
I approve now.

pfliu-nlp (approved):

PR #330, opened 2022-06-16
enhance correctcount metric to support multiple answers for Gaokao

pfliu-nlp:
For some tasks in Gaokao, there are multiple true answers formatted as `A/B`

### DIFF
--- a/explainaboard/metrics/accuracy.py
+++ b/explainaboard/metrics/accuracy.py
@@ -39,6 +39,19 @@ class CorrectCount(Accuracy):
     Calculate the absolute value of correct number
     """
 
+    def calc_stats_from_data(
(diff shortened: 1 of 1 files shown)


explainaboard/metrics/accuracy.py
```diff
@@ -39,6 +39,19 @@ class CorrectCount(Accuracy):
...
     """
 
+    def calc_stats_from_data(
+        self, true_data: list, pred_data: list, config: Optional[MetricConfig] = None
+    ) -> MetricStats:
+
+        return MetricStats(
+            np.array(
+                [
+                    (1.0 if y == x or y in str(x).split("/") else 0.0)
```
neubig, line 49: Instead of doing this splitting it might be better if the original dataset stored the multiple correct answers as a list?

PR #331, opened 2022-06-20
Significance testing efficiency

neubig:
This PR makes two major changes:
1. It switches to using Student's t-test for significance testing instead of the bootstrap when Student's t-test is applicable (i.e. when the corpus-level evaluation score is an average of the instance-level scores).
2. It batches the remaining unbatched parts of bootstrapping.
3. It turns plotting off by default in explainaboard_main unless `output_dir` is explicitly set.

This makes bucketing much faster (10x?) when Student's t-test is applicable, and somewhat faster when it is not (1.5x?)


pfliu-nlp (approved):
LGTM

pfliu-nlp (approved):

PR #332, opened 2022-06-21
added open_cached_file function to cache_api.py

ccatherinee:
I added an `open_cached_file` function to cache_api.py; this is done in order to better cache benchmark results (goes hand in hand with changes made in explainaboard_web). The `open_cached_file` function does the following:
opens the file only if it exists at get_cache_dir() + relative_path and is no older than timeout. Otherwise it will return None.


neubig (approved):
LGTM, thanks!

PR #333, opened 2022-06-22
fix-bugs-trainingset-dependent-features

pfliu-nlp:
#### Problem description:  explainaboard_web failed to deal with system outputs of generation tasks from named datasets -> training set dependent features are not supported -> SDK bug

This PR fixed this, but it is still very slow when dealing with system outputs from the test set of `cnn_dailymail`

### DIFF
--- a/explainaboard/processors/summarization.py
+++ b/explainaboard/processors/summarization.py
@@ -18,7 +18,7 @@ from explainaboard.processors.conditional_generation import (
     ConditionalGenerationProcessor,
 )
 from explainaboard.processors.processor_registry import register_processor
-import explainaboard.utils.feature_funcs
+from explainaboard.utils.feature_funcs import accumulate_vocab_from_samples
 from explainaboard.utils.py_utils import hash_dict
 from explainaboard.utils.typing_utils import unwrap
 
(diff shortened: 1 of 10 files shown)


neubig:
It looks like this is failing due to an upstream bug in DataLab when handling QA datasets. Could you fix that so tests pass?

Also, it'd be ideal if we could add a test that checks that training set dependent features are handled properly, as this is a pretty important feature.

pfliu-nlp:
> It looks like this is failing due to an upstream bug in DataLab when handling QA datasets. Could you fix that so tests pass?
> 

Yeah, I was interrupted by other things yesterday but will try to fix this ASAP.

> Also, it'd be ideal if we could add a test that checks that training set dependent features are handled properly, as this is a pretty important feature.

Yes! That's also something in my mind, I will add some tests.

pfliu-nlp:
> It looks like this is failing due to an upstream bug in DataLab when handling QA datasets. Could you fix that so tests pass?

Finally, found it's not the issue of DataLab (still because of independent feature bugs in other tasks). 
The reason why I can pass the unit tests locally is that the cache files in `.cache/explainaboard/stat/` are kept. So even if I set up a new `env` (but in the same server) I will still re-use the old caching files.
There is no fault with the current caching mechanism, I mention this just for a reminder in case we encounter these issues again in the future.

explainaboard/processors/summarization.py
```diff
@@ -106,26 +106,26 @@ def default_features(cls) -> feature.Features:
...
-                        dtype="float",
-                        description="the sample-level oracle position",
-                        is_bucket=True,
-                        bucket_info=feature.BucketInfo(
-                            method="bucket_attribute_specified_bucket_value",
-                            number=4,
-                            setting=(),
-                        ),
-                    ),
+                    # "oracle_score": feature.Value(
```
neubig, line 109:
```suggestion
                    # TODO: these are commented out because the implementation is currently 
                    #              too slow to have them be enabled by default
                    # "oracle_score": feature.Value(
```

explainaboard/processors/text_pair_classification.py
```diff
@@ -115,14 +115,14 @@ def default_metrics(
...
-        return {
-            'source_vocab': accumulate_vocab_from_samples(
-                samples, lambda x: x['text1'], unwrap(sys_info.source_tokenizer)
-            ),
-            'target_vocab': accumulate_vocab_from_samples(
-                samples, lambda x: x['text2'], unwrap(sys_info.target_tokenizer)
-            ),
-        }
+
+        source_vocab, source_vocab_rank = accumulate_vocab_from_samples(
```
neubig, line 119: I understand why you combined these together, but I think it'd be better to calculate the vocabularies separately. For example, there might be a cross-lingual text pair classification task where the source and target vocabs would be in different languages.

explainaboard/utils/feature_funcs.py
```diff
@@ -61,10 +61,12 @@ def accumulate_vocab_from_samples(
...
-    return {
-        "vocab": vocab,
-        "vocab_rank": vocab_rank,
-    }
+    # return vocab
+    return vocab, vocab_rank
+    # return {
+    #     "vocab": vocab,
+    #     "vocab_rank": vocab_rank,
+    # }
```
neubig, line 69:
```suggestion
    return vocab, vocab_rank
```

neubig (commented):
Looks good. But would it be possible to add a unit test as well?

neubig:
It seems that maybe using `cnn_dailymail` for unit tests is too heavy. Maybe we need to use a different dataset for this test, such as `conala`?

neubig (commented):
I approve other than `text_pair_classification.py` not using the target tokenizer for `text2`, which I think probably should be fixed. One other comment.

neubig (approved):

PR #334, opened 2022-07-05
Analysis levels

neubig:
This PR refactors how we specify which analyses we'd like to perform with the goal of (1) making the code cleaner particularly for tasks that perform analysis on multiple levels (e.g. NER with example-level and span-level), (2) allowing users to define custom features on multiple levels, and (3) making it possible to easily define analyses other than bucketing.

Current big changes include:

* Refactoring of how we specify features. Now features are directly associated with a feature function, which means that the function associated with features can be viewed in the same place as other information about the feature itself. See [text_classification.py](https://github.com/neulab/ExplainaBoard/blob/analysis_levels/explainaboard/processors/text_classification.py#L48) for an example.
* Split of feature definition into multiple levels, including over examples, spans, or any other thing that we'd like to perform analysis or show examples of. See [sequence_labeling.py](https://github.com/neulab/ExplainaBoard/blob/analysis_levels/explainaboard/processors/sequence_labeling.py#L37) for an example.
* Separation of specification of which analyses we perform from specification of features. This will give us the flexibility to specify analyses other than bucketing. See [text_classification.py](https://github.com/neulab/ExplainaBoard/blob/analysis_levels/explainaboard/processors/text_classification.py#L99) for an example.

This PR is basically done, but note that it is a PR into the `v0.11` branch, not `main`. As I think there is still significant work to be done before we could switch this over to being the main branch. Here are two milestones that I think still need to be finished:
* https://github.com/neulab/ExplainaBoard/milestone/2
* https://github.com/neulab/explainaboard_web/milestone/6


explainaboard/tests/artifacts/machine_translation/output_with_features.json
```diff
@@ -1,11 +1,21 @@
 {
   "metadata": {
     "custom_features": {
-      "num_capital_letters": {
-        "dtype": "float",
-        "description": "number of capital letters in the string",
-        "num_buckets": 4
+      "example": {
+        "num_capital_letters": {
+          "dtype": "float"
```
pfliu-nlp, line 6: we probably also want to provide an argument to take in customized feature functions in the future.

pfliu-nlp (approved):
finished the first-pass reviewing, which looks great.

explainaboard/info.py
```diff
@@ -147,21 +83,11 @@ def from_dict(cls, data_dict: dict) -> BucketPerformance:
...
-    def __len__(self):
-        return len(self.samples)
-
-
 @dataclass
 class Result:
-    overall: Optional[dict[str, Performance]] = None
+    overall: Optional[list[list[Performance]]] = None
     # {feature_name: {bucket_name: performance}}
-    fine_grained: Optional[dict[str, list[BucketPerformance]]] = None
```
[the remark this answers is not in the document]
neubig: Opened an issue for this: https://github.com/neulab/explainaboard_web/issues/245

explainaboard/tests/artifacts/machine_translation/output_with_features.json
neubig: I agree, there's an issue open for this: https://github.com/neulab/ExplainaBoard/issues/338

PR #335, opened 2022-07-09
support multiple true answers

pfliu-nlp:
* Re-format  `answers` of `cloze-generative` task: from str to list 
* extend correctcount metric


neubig (approved):
LGTM!

PR #339, opened 2022-07-10
fix typo in kg

pfliu-nlp:


pfliu-nlp:
Two small fixes for the kg task
* add hit10 metric
* fix typo of feature description

PR #343, opened 2022-07-13
try to make datalab dataloader task-specific

pfliu-nlp:
This PR tries to make datalab-style data loader of explainaboard dataset-specific by specifying  `field column name` instead  of `field name`
for example, for xnli task,
* field column name: `text1_column`, `text2_column`, `label`
* column name: `sentence1`, `sentence2`, `gold_label`

Caveat: machine translation datasets from datalab haven't been considered.


neubig (approved):
Looks great, thanks a bunch! Maybe we can update the minor version number of the SDK, and then do an update of `explainaboard_web` so we can fix this in the web interface?

pfliu-nlp:
>"Maybe we can update the minor version number of the SDK, and then do an update of explainaboard_web so we can fix this in the web interface?"

Sounds good!

Issue #337, opened 2022-07-09
Add confusion matrix analysis


neubig:
As an example of another analysis type enabled by the new analysis levels, we can add confusion matrices.

neubig:
Done

PR #345, opened 2022-07-17
Feature combination count analysis

neubig:
This adds the ability to count the number of times combinations of particular features appear in a system output. One of the main use cases for this will be calculating confusion matrices between true and predicted tags, and I tried to add this to every task where this is relevant.

Fixes: https://github.com/neulab/ExplainaBoard/issues/337

### DIFF
--- a/explainaboard/analysis/analyses.py
+++ b/explainaboard/analysis/analyses.py
@@ -177,6 +177,63 @@ class BucketAnalysis(Analysis):
 
         return BucketAnalysisResult(self.feature, bucket_performances)
 
+@dataclass
+class ComboCountAnalysisResult(AnalysisResult):
+    features: tuple
+    combo_counts: dict[tuple, int] | None = None
+    _type: Optional[str] = None
+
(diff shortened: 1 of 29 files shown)


explainaboard/analysis/analyses.py
```diff
@@ -177,6 +177,63 @@ def perform(
...
+            self,
+            cases: list[AnalysisCase],
+            metrics: list[Metric],
+            stats: list[MetricStats],
+            conf_value: float,
+    ) -> AnalysisResult | None:
+        if len(cases) == 0 or any((x not in cases[0].features) for x in self.features):
+            return None
+        combo_counts: dict[tuple, int] = {}
+        for case in cases:
```
pfliu-nlp, line 241: It would be nice if some docstrings could be added here.

neubig: Thanks for the comment @pfliu-nlp ! I've resolved it.

pfliu-nlp (approved):

PR #347, opened 2022-07-20
Tabular data

neubig:
This PR adds support for two new varieties of task:
* Tabular Classification: classification based on tabular data
* Tabular Regression: regression based on tabular data


pfliu-nlp (approved):
looks good

PR #348, opened 2022-07-22
Initial version of rank-flipping meta-/quantitative-analysis

hwidjaja:


explainaboard/meta_analyses/rank_flipping.py
```diff
@@ -0,0 +1,185 @@
+class RankFlippingMetaAnalysis:  # (can inherit from an abstract MetaAnalysis class)
+    def __init__(self, model1_report, model2_report):
```
pfliu-nlp, line 2: it would be nice if we could add type info to these arguments, for example, `model1_report:dict`

PR #351, opened 2022-07-26
introduce unwrap_or and remove Metric._get_config

odashi:
This pr introduces `typing_utils.unwrap_or`, which provides a similar semantics to Rust's `Optional.unwrap_or`.
This pr also removes `Metric._get_config` since this function does nothing else than `unwrap_or`.


neubig (approved):
LGTM, thanks!

PR #352, opened 2022-07-26
Refactor typing_utils.narrow

odashi:
This pr introduces some small changes in `typing_utils.narrow`:

- Write a complete documentation.
- Use `__name__` in the error string rather than the `str` representation.
- Change the argument order to `narrow(type, obj)` to keep consistency with `typing.cast`


odashi:
Some tricky error occurred around `cast`, investigating the cause.

odashi:
`cast` does not work with TypeVar.
https://github.com/python/mypy/issues/6440

As the type inference can be performed correctly without `cast` (because of the `isinstance` narrowing), simply avoiding `cast` may be the right choice for now.

neubig (approved):
LGTM, thanks!

PR #354, opened 2022-07-27
Refactor NLG metrics

neubig:
This is an attempt to refactor NLG metrics to make them more general and less verbose, please take a look!


explainaboard/metrics/nlg_meta_evaluation.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    threshold: float = 25
+
+    def to_metric(self):
+        return KtauCorrelation(self)
+
+
+class KtauCorrelation(CorrelationMetric):
+    def aggregate_stats(self, stats: MetricStats) -> np.ndarray:
```
pfliu-nlp, line 148: it seems this function could be removed.

pfliu-nlp (approved):
looks good. Do you think all current implementations are kinda WMT dataset oriented? We could make it generalized gradually in future PR.

neubig:
> Do you think all current implementations are kinda WMT dataset oriented? We could make it generalized gradually in future PR.

Yeah, somewhat. I wanted to fix the above issue of generality before merging the current PR, but we can also generalize more in the future.

PR #355, opened 2022-07-27
Add human eval metrics

pfliu-nlp:
This PR implements the following features:
* added human evaluation metrics without actual human evaluated scores
* explored `likert`-based human evaluation first and defined a simple config for it (other methods could be introduced gradually)
* introduce multiple annotation features and agreement calculation

[Example](https://github.com/neulab/ExplainaBoard/blob/097d300fb52c2902ff14fe4b9a691bddf65b3ff8/explainaboard/tests/test_summarization.py#L103)
```python

        metadata = {
            "task_name": TaskType.summarization.value,
            "dataset_name": "cnndm",
            "metric_configs": [
                LikertScoreConfig(
                    name="LikertScore_fluency",
                    aspect="fluency",
                    n_annotators=2,
                    categories=5,
                    external_stats=np.array([[2, 2], [1, 1], [3, 3]]),
                )
            ],
        }

        processor = get_processor(TaskType.summarization.value)

        sys_info = processor.process(metadata, data)


```
Performance:
```
{'LikertScore_fluency': Performance(metric_name='LikertScore_fluency', value=2.0, confidence_score_low=1.0, confidence_score_high=3.0, agreement=1.0)}
```


neubig:
I'm a bit confused about how exactly this code fits in a workflow of using ExplainaBoard for some purpose. For instance, it's relatively obvious how you use the other metrics (you specify them when you run the CLI code or through the web interface), but it's not clear what you do for these human evaluation metrics. Could you add a little bit of documentation so this is more clear? Thanks!

pfliu-nlp:
> 

The expected use story is that:  when developers submit system outputs using ExplainaBoard Web, they are able to choose the human evaluation metrics from the drop-down box.

To achieve this goal, Explainaboard SDK should support human evaluation metrics.

pfliu-nlp:
hi, @neubig  modified a little bit based on our discussion.

neubig (approved):

neubig:
Looks good, thanks!

PR #358, opened 2022-07-28
Move supported_file_types to Loader

odashi:
This pr adds `Loader.supported_file_types` and removes the same implementation in `loader_registry` since this functionality should belong to `Loader`.

This also adds `loader_registry.get_loader_class` to obtain the Loader class which is required to use the Loader's class methods without initialization.

I guess `loader_registry.get_custom_dataset_loader` is also not needed: it can be replaced by simply initializing the class returned by `get_loader_class`. I will propose another pr to fix this.

### DIFF
--- a/explainaboard/loaders/loader.py
+++ b/explainaboard/loaders/loader.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


neubig (approved):
LGTM, thanks!

PR #359, opened 2022-07-28
Remove initialization functions form loader_registry

odashi:
Blocked by #358 

This pr replaces following functions in `loader_registry` with alternatives:

- `get_custom_datsaet_loader(task, ...)` -> `get_loader_class(task)(...)`
- `get_datalab_loader(task, ...)` -> `get_loader_class(task).init_datalab(...)`

These functions originally perform two things:

1. locate the loader class using the first argument
2. initialize the loader class using the rest of arguments

Considering the semantics of `loader_registry`, performing both at this module seems like overstepping, and it should focus on the step 1.
The step 2. is the responsibility of the loader itself, and can be performed separately after obtaining the loader class.


neubig (approved):
I approve as-is, but also for consideration: maybe we could have parallel functions `from_datalab()` (currently `init_datalab()`) and `from_local_file()` just to make the two options explicit. Either way is fine with me.

odashi:
@neubig Yeah, and if we get more functions to initialize Loader, it would be better to separate them to another module (such as `loader_utils` or something else with more specific name).

I also guessed `from_something` is more consistent name than `init_something`. will fix it in this pr.

neubig (approved):
Looks good, I approve.

PR #364, opened 2022-07-30
Add text_writer context manager

odashi:
This pr adds `utils.io_utils.text_writer(file)` context manager, and replace the behavior to make reports in the main module to this functionality. Unlike the current implemenation, this function correctly close the file when some exception raised during generating the report file.


neubig (approved):
LGTM!

PR #365, opened 2022-07-30
declare customized features from config

pfliu-nlp:
This PR aims to make customize features from datalab be declared from a [config file](https://github.com/neulab/ExplainaBoard/blob/add_customized_features_from_config/explainaboard/resources/dataset_custom_features.json) and apply this functionality to open-domain QA task.

`dataset_custom_features.json`
```json
{
  "sst2": {
    "label": {
      "dtype": "string",
      "description": "the true label",
      "num_buckets": 2
    }
  },
  "natural_questions_comp_gen": {
    "question_types": {
      "dtype": "string",
      "description": "the question type",
      "num_buckets": 5
    }
  }
}

```


explainaboard/loaders/file_loader.py
```diff
@@ -526,6 +527,15 @@ def load_raw(
         self, data: str | DatalabLoaderOption, source: Source
     ) -> FileLoaderReturn:
         config = narrow(DatalabLoaderOption, data)
+
+        # load customized features from global config files
+        customized_features_from_config = get_custmomized_features()
```
neubig, line 532: typo: should be "customized"

explainaboard/loaders/file_loader.py
```diff
@@ -526,6 +527,15 @@ def load_raw(
...
+        if config.dataset in customized_features_from_config.keys():
```
neubig, line 533:
```suggestion
        if config.dataset in customized_features_from_config:
```

explainaboard/loaders/file_loader.py
```diff
@@ -547,6 +557,14 @@ def load_raw(
 
         # Infer metadata from the dataset
         metadata = FileLoaderMetadata()
+        # load customized features from global config files
+        if config.dataset in customized_features_from_config.keys():
```
neubig, line 561:
```suggestion
        if config.dataset in customized_features_from_config:
```

explainaboard/resources/dataset_custom_features.json
```diff
@@ -0,0 +1,16 @@
+{
```
neubig, line 1: Could you please add the features for `sumeval2022` as well?

neubig (commented):
Just a few suggestions.

explainaboard/loaders/file_loader.py
pfliu-nlp: I think `if config.dataset in customized_features_from_config.keys()` is correct here.

explainaboard/loaders/file_loader.py
pfliu-nlp: I think `if config.dataset in customized_features_from_config.keys()` is correct here.

explainaboard/loaders/file_loader.py
neubig: Sorry, I wasn't clear enough on this. I think they're both functionally identical, but:
1. writing `.keys()` is more verbose, and
2. I believe `if x in y.keys()` is `O(len(y))`, where `if x in y` is `O(1)`, as the former returns a view of the keys as a list: https://www.w3schools.com/python/ref_dictionary_keys.asp

So I'd suggest not writing `.keys()`

neubig: Actually I was wondering about "2." so I researched a bit more, and it seems that starting in Python 3.x both are `O(1)`: https://stackoverflow.com/a/17539423

But still, not writing `.keys()` is less verbose, so I'd recommend it either way.

pfliu-nlp: oh, I see. Thanks for the patient explanation!

neubig (approved):
LGTM.

PR #369, opened 2022-08-01
add get_metric_config_class

odashi:
This pr adds `get_metric_config_class` to simplify the behavior of the registry. The function returns the `MetricConfig` class associated to the given name.


neubig (approved):
LGTM!

PR #372, opened 2022-08-02
ExplainaBoard V0.11

neubig:
This is a major refactoring of ExplainaBoard to:
1. Introduce the concept of "analysis levels", where analysis can be performed over examples, spans, tokens, etc. separately. In order to implement this, the code is refactored to first generate "analysis cases" for each of these levels, then perform analysis over the analysis cases.
2. Separate definition of features from definition of analyses. This allows for multi-feature analyses (like confusion matrices, etc.). Features can also be defined directly through feature functions instead of relying on separately implemented functions.


pfliu-nlp (approved):

explainaboard/analysis/analyses.py
```diff
@@ -0,0 +1,317 @@
...
+
+    name: str
+
+    def print(self):
+        raise NotImplementedError
+
+    @staticmethod
+    def from_dict(dikt):
+        type = dikt.pop('cls_name')
+        if type == 'BucketAnalysisResult':
```
odashi, line 35: This is a subclass specific behavior, should be saparated from the base class.

Consider implementing an external dispatcher function:
```python
def load_analysis_result_from_dict(d) -> AnalysisResult:
    cls_name = d['cls_name']
    if cls_name == 'BucketAnalysisResult':
        return BucketAnalysisResult.from_dict(d)
    elif
    ...

class AnalysisResult:
    @classmethod
    @abstractmethod
    def from_dict(cls, d):
        pass

class BucketAnalysisResult:
    @classmethod
    def from_dict(cls, d):
        ...
```

explainaboard/analysis/analyses.py
```diff
@@ -0,0 +1,317 @@
...
+        :param stats: The statistics calculated by each metric.
+        :conf_value: In the case that any significance analysis is performed, the
+          confidence level.
+        """
+        raise NotImplementedError
+
+    @staticmethod
+    def from_dict(dikt: dict):
+        type = dikt.pop('cls_name')
+        if type == 'BucketAnalysis':
```
odashi, line 86: ditto.

explainaboard/analysis/case.py
```diff
@@ -0,0 +1,82 @@
...
+    sample_id: int
+    features: dict
+
+    def __post_init__(self):
+        if isinstance(self.sample_id, str):
+            raise ValueError
+
+    @classmethod
+    def dict_conv(cls, k: str, v: dict):
+        return v
```
odashi, line 23: Add description: I didn't understand what this function is.

explainaboard/analysis/case.py
```diff
@@ -0,0 +1,82 @@
...
+    :param predicted_label: The label that is predicted
+    """
+
+    true_label: str
+    predicted_label: str
+
+
+@dataclass
+class AnalysisCaseCollection:
+    interval: tuple
```
odashi, line 78: tuple of any?

explainaboard/analysis/feature.py
```diff
@@ -1,39 +1,21 @@
...
-
-    def __post_init__(self):
-        self._type: str = self.__class__.__name__
+from typing import Optional
 
 
 def is_dataclass_dict(obj):
     """
-    this function is used to judge if obj is a dictionary with a key `_type`
+    this function is used to judge if obj is a dictionary with a key `cls_name`
```
odashi, line 11: It looks the description does not explain the correct behavior of this function.

explainaboard/analysis/feature.py
```diff
@@ -1,39 +1,21 @@
...
     :param obj: a python object with different potential type
     :return: boolean variable
     """
     if (
         not isinstance(obj, dict)
-        or '_type' not in obj.keys()
-        or obj['_type'] not in FEATURETYPE_REGISTRY.keys()
+        or 'cls_name' not in obj.keys()
```
odashi, line 17: It looks better to refactor the whole function.
```suggestion
def is_feature_type_dict(obj: Any) -> bool:
    return (
        isinstance(obj, dict)
        and 'cls_name' in obj
        and obj['cls_name'] in REATURETYPE_REGISTRY
    )
```

explainaboard/analysis/feature_funcs.py
```diff
@@ -1,14 +1,34 @@
...
 
+from explainaboard.info import SysOutputInfo
 from explainaboard.utils import basic_words
 from explainaboard.utils.logging import progress
 from explainaboard.utils.tokenizer import Tokenizer
+from explainaboard.utils.typing_utils import unwrap
+
+
+def _get_tokens(
+    sys_info: SysOutputInfo, text: str | list[str], side: str = 'source'
```
odashi, line 16: Why is the "source" set as the default?

explainaboard/analysis/feature_funcs.py
```diff
@@ -1,14 +1,34 @@
...
+) -> list[str]:
+    if isinstance(text, list):
+        if side != 'none':
```
odashi, line 19: Is there any reason to take care about the value of `side` here? If no, simply ignoring this value looks better.

explainaboard/analysis/feature_funcs.py
```diff
@@ -65,36 +85,44 @@ def accumulate_vocab_from_samples(
 
 
 def feat_freq_rank(
-    existing_features: dict,
-    statistics: Any,
-    text_from_sample: Callable,
-    tokenizer: Tokenizer,
-):
+    sys_info: SysOutputInfo,
```
odashi, line 88: It looks `feat_freq_rank`, `feat_num_oov`, and `feat_length_freq` functions need to avoid `_get_tokens`. Arguments `sys_info`, `text`, `side` are used to only invoke `_get_tokens`. If `_get_tokens` is invoked before calling these functions, they could become much simpler.

explainaboard/analysis/feature_funcs.py
```diff
@@ -111,3 +139,19 @@ def cap_feature(s):
         return "first_caps"
     else:
         return "not_first_caps"
+
+
+def relative_position(sys_info: SysOutputInfo, text: str, word: str):
+    tokens = unwrap(sys_info.source_tokenizer)(text).strs
+    if word not in tokens:
+        return 0
```
odashi, line 147: Is there any meaning of this particular value?
Since 0 seems to be also representing the first position, I think returning `None` would be better.

explainaboard/analysis/feature_funcs.py
```diff
@@ -111,3 +139,19 @@ def cap_feature(s):
...
+    else:
+        return float(tokens.index(word)) / len(tokens)
+
+
+def absolute_position(sys_info: SysOutputInfo, text: str, word: str):
+    tokens = unwrap(sys_info.source_tokenizer)(text).strs
+    if word not in tokens:
+        return 0
```
odashi, line 155: ditto.

explainaboard/metrics/registry.py
```diff
@@ -30,3 +32,12 @@ def get_metric_config_class(name) -> type[MetricConfig]:
...
     raise ValueError(f'Invalid Metric {name}')
+
+
+def metric_config_from_dict(dikt: dict):
+    type = dikt.pop('cls_name')
+    config_cls = get_metric_config_class(type)
+    field_names = set(f.name for f in dataclasses.fields(config_cls))
+    return config_cls(
+        **{k: config_cls.dict_conv(k, v) for k, v in dikt.items() if k in field_names}
+    )
```
odashi, line 43: It looks implemnting `MetricConfig.from_dict(d)` is better than manually manipulating the members.

odashi (commented):
I couldn't guarantee the quality of my review since the change is too large to understand.

explainaboard/loaders/qa_open_domain.py
```diff
@@ -31,16 +31,16 @@ def default_dataset_file_loaders(cls) -> dict[FileType, FileLoader]:
...
                 ]
             ),
             FileType.datalab: DatalabFileLoader(
                 [
                     FileLoaderField("question_column", target_field_names[0], str),
                     FileLoaderField("answers_column", target_field_names[1], list),
-                    FileLoaderField(
-                        "question_types_column", target_field_names[2], list
-                    ),
+                    # FileLoaderField(
```
odashi, line 41: Could you remove commented code?

explainaboard/processors/extractive_qa.py
```diff
@@ -22,77 +28,75 @@ class QAExtractiveProcessor(Processor):
...
+            )
+            for x in continuous_features
+        ]
+
+        return [
+            AnalysisLevel(
+                name='example',
+                features=features,
+                metric_configs=self.default_metrics(),
+                analyses=cast(List[Analysis], analyses),
```
odashi, line 93: Any reason to use `List`?
(ditto for others)

explainaboard/processors/qa_open_domain.py
```diff
@@ -22,72 +28,69 @@ class QAOpenDomainProcessor(Processor):
...
+        features = {
+            "question": feature.Value("string"),
+            # "question_types": feature.Sequence(feature=feature.Value("string")),
+            "answers": feature.Sequence(feature=feature.Value("string")),
+            "question_length": feature.Value(
+                dtype="float",
+                description="context length in tokens",
+                func=lambda info, x, c: count_tokens(info, x['question']),
+            ),
+            # "question_type": feature.Value(
```
odashi, line 41: Could you remove commented code? Since we are using CVS, there is almost no reason to keep unused code.

explainaboard/processors/summarization.py
```diff
@@ -60,135 +61,66 @@ class SummarizationProcessor(ConditionalGenerationProcessor):
...
+                dtype="float",
+                description="coverage",
+                func=lambda info, x, c: c.features['sum_attributes']["attr_coverage"],
             ),
+            "attr_novelty": feature.Value(
+                dtype="float",
+                description="novelty",
+                func=lambda info, x, c: c.features['sum_attributes']["attr_novelty"],
+            ),
+            # "oracle_score": feature.Value(
```
odashi, line 95: same

explainaboard/analysis/feature_funcs.py
neubig: I agree. I think it might be a good idea to refactor this in a later PR, but maybe not here just to limit the scope of changes from getting larger.

explainaboard/analysis/feature_funcs.py
neubig: I agree that 0 is not good. `None` would be best, but it would break our downstream processing, so I set it to "-1" for now.

explainaboard/metrics/registry.py
neubig: I agree, although I'd suggest we save this for a future PR.

explainaboard/processors/extractive_qa.py
neubig: This was due to a peculiarity of Python 3.8, but this code has been removed so it's no long relevant.

odashi (commented):
I couldn't give my approval since it is hard to review the entire appropriateness. Feel free to push this change with the maintainer's responsibility.

neubig:
OK, thanks, and I understand. This is a big change that necessitated lots of code changes, and I'll also strive to do more granular PR commits in the future.

PR #378, opened 2022-08-10
Fix missing resources error

pfliu-nlp:
Previously, the config file `dataset_custom_features.json` can not be downloaded when performing `pip install explainaboard`


pfliu-nlp:
I merged it to test if pip could work with having `resources` folder.

jinlanfu:
It looks good!

jinlanfu (approved):
Nice!

PR #380, opened 2022-08-12
Modify analyses to be flat list

neubig:
This modifies v0.11 so that analyses are flat lists, instead of a hierarchical list with one analysis per level. The biggest change is splitting the specification of the features and metrics for each analysis level, and the analyses themselves into different functions:

```
    @abc.abstractmethod
    def default_analysis_levels(self) -> list[AnalysisLevel]:
        """Returns a list of analysis levels, indicating analyses that can be
        applied to different views of the higher-level example. For instance, a task may
        perform 'example'-level analysis, and 'token'-level analysis in which case this
        list would have one level for each."""
        ...

    @abc.abstractmethod
    def default_analyses(self) -> list[Analysis]:
        """Returns the analyses to be performed."""
        ...
```

### DIFF
--- a/explainaboard/processors/aspect_based_sentiment_classification.py
+++ b/explainaboard/processors/aspect_based_sentiment_classification.py
@@ -79,7 +79,6 @@ class AspectBasedSentimentClassificationProcessor(Processor):
 
     def default_analyses(self) -> list[Analysis]:
         features = self.default_analysis_levels()[0].features
-        continuous_features = [k for k, v in features.items() if v.dtype == 'float32']
         # Create analyses
         analyses: list[Analysis] = [
             BucketAnalysis(
@@ -95,15 +94,8 @@ class AspectBasedSentimentClassificationProcessor(Processor):
                 features=("true_label", "predicted_label"),
             ),
         ]
-        for x in continuous_features:
-            analyses.append(
-                BucketAnalysis(
-                    level="example",
-                    description=features[x].description,
-                    feature=x,
-                    method="continuous",
-                )
-            )
+        # Continuous features
+        analyses.extend(super().default_analyses())
         return analyses
 
     @classmethod
--- a/explainaboard/processors/aspect_based_sentiment_classification.py
+++ b/explainaboard/processors/aspect_based_sentiment_classification.py
@@ -95,7 +95,7 @@ class AspectBasedSentimentClassificationProcessor(Processor):
             ),
         ]
         # Continuous features
-        analyses.extend(super().default_analyses())
+        analyses.extend(self.continuous_feature_analyses())
         return analyses
 
     @classmethod
--- a/explainaboard/processors/aspect_based_sentiment_classification.py
+++ b/explainaboard/processors/aspect_based_sentiment_classification.py
@@ -15,7 +15,6 @@ from explainaboard.metrics.metric import MetricConfig
 from explainaboard.processors.processor import Processor
 from explainaboard.processors.processor_registry import register_processor
 from explainaboard.utils.spacy_loader import get_named_entities
-from explainaboard.utils.typing_utils import unwrap
 
 
 @register_processor(TaskType.aspect_based_sentiment_classification)
@@ -80,9 +79,7 @@ class AspectBasedSentimentClassificationProcessor(Processor):
 
     def default_analyses(self) -> list[Analysis]:
         features = self.default_analysis_levels()[0].features
-        continuous_features = [
-            k for k, v in features.items() if ('float' in unwrap(v.dtype))
-        ]
+        continuous_features = [k for k, v in features.items() if v.dtype == 'float32']
         # Create analyses
         analyses: list[Analysis] = [
             BucketAnalysis(
--- a/explainaboard/processors/aspect_based_sentiment_classification.py
+++ b/explainaboard/processors/aspect_based_sentiment_classification.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import cast, List
-
 from explainaboard import TaskType
 from explainaboard.analysis import feature
 from explainaboard.analysis.analyses import (
@@ -26,7 +24,7 @@ class AspectBasedSentimentClassificationProcessor(Processor):
     def task_type(cls) -> TaskType:
         return TaskType.aspect_based_sentiment_classification
 
-    def default_analyses(self) -> list[AnalysisLevel]:
+    def default_analysis_levels(self) -> list[AnalysisLevel]:
         features: dict[str, FeatureType] = {
             "aspect": feature.Value(
                 dtype="string",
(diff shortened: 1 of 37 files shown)


explainaboard/processors/conditional_generation.py
```diff
@@ -156,31 +145,54 @@ def default_analyses(self) -> list[AnalysisLevel]:
...
+                BucketAnalysis(
+                    level="example",
+                    description=examp_features[x].description,
+                    feature=x,
+                    method="continuous",
+                )
+                for x in examp_continuous_features
+            ]
+        )
+        tok_features = self.default_analysis_levels()[1].features
```
odashi, line 179: Store return values of `default_analysis_levels` to avoid calling it twice.

explainaboard/processors/conditional_generation.py
```diff
@@ -156,31 +145,54 @@ def default_analyses(self) -> list[AnalysisLevel]:
...
         return [
             AnalysisLevel(
                 name='example',
                 features=examp_features,
                 metric_configs=self.default_metrics(level='example'),
-                analyses=cast(List[Analysis], examp_analyses),
             ),
             AnalysisLevel(
-                name='tok',
+                name='token',
```
odashi, line 156: This looks a breaking change. Be careful to check if this change won't affect anything.

explainaboard/processors/language_modeling.py
```diff
@@ -125,34 +114,57 @@ def default_analyses(self) -> list[AnalysisLevel]:
...
-                analyses=cast(List[Analysis], tok_analyses),
+                metric_configs=self.default_metrics(level='token'),
             ),
         ]
 
+    def default_analyses(self) -> list[Analysis]:
+        analyses: list[Analysis] = []
+        examp_features = self.default_analysis_levels()[0].features
+        examp_continuous_features = [
+            k for k, v in examp_features.items() if ('float' in unwrap(v.dtype))
```
odashi, line 135: This looks dangerous: consider allowing only the exact type. (same for others)

```suggestion
            k for k, v in examp_features.items() if (v.dtype == 'float')
```

explainaboard/processors/language_modeling.py
```diff
@@ -125,34 +114,57 @@ def default_analyses(self) -> list[AnalysisLevel]:
...
+            [
+                BucketAnalysis(
+                    level="token",
+                    description=tok_features[x].description,
+                    feature=x,
+                    method="continuous",
+                )
+                for x in tok_continuous_features
+            ]
+        )
```
odashi, line 165: Several times the `Analysis` object is generated via multiple comprehension. Using a usual for-loop would be much simpler and faster. (same for others)
```suggestion
        for feature_name, feature_value in self.default_analysis_levels()[1].features:
            if feature_name == "tok_log_prob":
                break
            if "float" in unwrap(v.dtype):
                break
            analyses.append(
                BucketAnalysis(
                    level="token",
                    descritpion=feature_value.description,
                    method="continuous",
                )
            )
```

explainaboard/processors/qa_multiple_choice.py
```diff
@@ -77,25 +76,31 @@ def default_analyses(self) -> list[AnalysisLevel]:
...
-                analyses=cast(List[Analysis], analyses),
             )
         ]
 
+    def default_analyses(self) -> list[Analysis]:
+        features = self.default_analysis_levels()[0].features
+        continuous_features = [
+            k for k, v in features.items() if ('float' in unwrap(v.dtype))
+        ]
+        analyses: list[Analysis] = [
```
odashi, line 93:
```suggestion
        return [
```

explainaboard/processors/qa_open_domain.py
```diff
@@ -69,25 +68,31 @@ def default_analyses(self) -> list[AnalysisLevel]:
...
-                analyses=cast(List[Analysis], analyses),
             )
         ]
 
+    def default_analyses(self) -> list[Analysis]:
+        features = self.default_analysis_levels()[0].features
+        continuous_features = [
+            k for k, v in features.items() if ('float' in unwrap(v.dtype))
+        ]
+        analyses: list[Analysis] = [
```
odashi, line 85:
```suggestion
        return [
```

explainaboard/visualizers/performance_gap.py
```diff
@@ -20,24 +20,23 @@ def get_pairwise_performance_gap(
...
-            for fgr_buk, fgr1_buk, fgr2_buk in zip(
-                fgr_buks.bucket_performances,
-                fgr1_buks.bucket_performances,
-                fgr2_buks.bucket_performances,
+    for fgr_lev, fgr1_lev, fgr2_lev in zip(fgr, fgr1, fgr2):
+        if not isinstance(fgr_lev, BucketAnalysisResult):
+            continue
+        fgr_buks, fgr1_buks, fgr2_buks = (
+            cast(BucketAnalysisResult, x) for x in (fgr_lev, fgr1_lev, fgr2_lev)
+        )
```
odashi, line 28: This casting should not be necessary since `isinstance` in the previous statement already narrows the type.

explainaboard/processors/conditional_generation.py
neubig: This PR is modifying the v0.11 branch, which isn't yet merged into the main branch, so I think it's OK.
If this were already merged into the main branch it would indeed break backward compatibility with previously generated reports so we'd need to be a bit more careful.

explainaboard/processors/aspect_based_sentiment_classification.py
```diff
@@ -72,37 +70,44 @@ def default_analyses(self) -> list[AnalysisLevel]:
             ),
         }
 
+        return [
```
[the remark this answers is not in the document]
neubig: It's used in `processor.py`

explainaboard/processors/conditional_generation.py
odashi: Oh okay, but could you avoid merging to any non-main branches? Merging to arbitrary branches makes us hard to track the actual progress of the development.

explainaboard/processors/processor.py
```diff
@@ -46,10 +47,26 @@ def default_analysis_levels(self) -> list[AnalysisLevel]:
...
         ...
 
-    @abc.abstractmethod
     def default_analyses(self) -> list[Analysis]:
-        """Returns the analyses to be performed."""
-        ...
+        """Returns the analyses to be performed.
+
+        By default it performs analysis over
+        all continuous features specified in the analysis levels."""
```
odashi, line 54: This is not a good idea for several reasons:

- No reason to decide this as the default behavior.
- Users couldn't guess this behavior correctly from the responsibility of "processor".
- Changing this behavior affects every subclasses, which are not always managed by us.

If you need to implement some "common features", introduce an additional utility rather than directly implementing it to the base method.

Not this:
```python
class Base:
  def foo(self):
    one()
    two()  # these are not considered as the Base's functionality

class Derived:
  def foo(self):
    super().foo()
    three()    
```

Implement this instead:
```python
class Base(ABC):
  @abstractmethod
  def foo(self):
    pass

def do_common():  # better to be placed outside Base if possible
  one()
  two()

class Derived:
  def foo(self):
    do_common()
    three()
```

explainaboard/processors/conditional_generation.py
neubig: That makes sense. The justification behind doing it here is because I wanted to make it easier for you to review these changes independently of the very large PR that we're currently waiting to merge here (once you approve of it of course): https://github.com/neulab/ExplainaBoard/pull/372

explainaboard/processors/processor.py
neubig: OK, this makes sense. It'll make the code a slight bit more verbose when using the default behavior, but I think that's a small price to pay for clarity.

Quick question: why is it better to have "do_common" placed outside of Base?

odashi: > why is it better to have "do_common" placed outside of Base?
This depends on the situation. In this case, I observed that the "common behavior" is filtering a float features to create a continuous analysis, and the behavior itself is almost independent from the processor.

explainaboard/processors/conditional_generation.py
odashi: Ah sure. It's somewhat unfortunate since GitHub lacks feature to compare different base commit than main.

odashi (approved):

Issue #346, opened 2022-07-17
Visualization for confusion matrices


neubig:
Currently the visualization scripts will not output visualizations for CombinationCountAnalysis, but they should.

PR #384, opened 2022-08-12
Confusion matrix visualization

neubig:
This adds rudimentary visualization for confusion matrices to the offline visualization tool. These will be output automatically when using the `draw_charts` utility (which used to be called the `draw_hist` utility)

true_label_predicted_label_combo_0
true_label_predicted_label_combo_1

Fixes https://github.com/neulab/ExplainaBoard/issues/346

Blocked by https://github.com/neulab/ExplainaBoard/pull/372


pfliu-nlp (approved):
looks good

neubig:
@odashi this is no longer blocked, if you'd like to take a look before we merge.

explainaboard/visualizers/draw_charts.py
```diff
@@ -0,0 +1,265 @@
...
+        for feats, count in unwrap(combo_result.combo_counts):
+            for feat, featmap in zip(feats, feature_maps):
+                featmap[feat] = featmap.get(feat, 0) + count
+    # sort in descending order of frequency
+    sorted_names = [
+        [v[0] for v in sorted(x.items(), key=lambda y: -y[1])] for x in feature_maps
+    ]
+    first_set = set(sorted_names[0])
+    if all([set(x) == first_set for x in sorted_names]):
+        sorted_names = [sorted_names[0] for _ in sorted_names]
```
odashi, line 50: The intention of this is somewhat unclear to me: why do we need to apply this restriction? and what is the reason to choose the first list of names rather than considering the whole frequency?

explainaboard/visualizers/draw_charts.py
```diff
@@ -0,0 +1,265 @@
...
+    :param output_dir:
+    :return:
+    """
+
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+
+    # TODO(gneubig): This should get the system name from inside the report
+    if sys_names is None:
+        sys_names = [os.path.basename(x).replace('.json', '') for x in reports]
```
odashi, line 153: Use `splitext` rather than ad-hoc code.
```suggestion
        sys_names = [os.path.splitext(os.path.basename(x))[0] for x in reports]
```

explainaboard/visualizers/draw_charts.py
```diff
@@ -0,0 +1,265 @@
...
+            title=None,
+            xticklabels=overall_metric_names,
+            ylabel='metric value',
+        )
+
+    # --- analysis results
+    analysis_results: list[list[AnalysisResult]] = [
+        unwrap(x.results.analyses) for x in report_info
+    ]
+    if any([len(x) != len(analysis_results[0]) for x in analysis_results]):
```
odashi, line 198: ditto
```suggestion
    if any(len(x) != len(analysis_results[0]) for x in analysis_results):
```

explainaboard/visualizers/draw_charts.py
```diff
@@ -0,0 +1,265 @@
...
+        raise ValueError(
+            f'mismatched number of analyses: {[len(x) for x in analysis_results]}'
+        )
+
+    # Bucket performance: feature name, for example, sentence length
+    for analysis_result in progress(zip(*analysis_results)):
+        if any([x.name != analysis_result[0].name for x in analysis_result]):
```
odashi, line 205: ditto
```suggestion
        if any(x.name != analysis_result[0].name for x in analysis_result):
```

explainaboard/visualizers/draw_charts.py
```diff
@@ -0,0 +1,265 @@
...
+            raise ValueError(
+                f'mismatched analyses: {[x.name for x in analysis_result]}'
+            )
+
+        if all([isinstance(x, BucketAnalysisResult) for x in analysis_result]):
```
odashi, line 210:
```suggestion
        if all(isinstance(x, BucketAnalysisResult) for x in analysis_result):
```

explainaboard/visualizers/draw_charts.py
```diff
@@ -0,0 +1,265 @@
...
+            plot_buckets(analysis_result, output_dir, sys_names)
+        elif all([isinstance(x, ComboCountAnalysisResult) for x in analysis_result]):
```
odashi, line 212:
```suggestion
        elif all(isinstance(x, ComboCountAnalysisResult) for x in analysis_result):
```

explainaboard/visualizers/draw_charts.py
```diff
@@ -0,0 +1,265 @@
...
+        '--reports',
+        type=str,
+        required=True,
+        nargs="+",
+        help="the directories of reports. Multiple one should be separated by space, "
+        "for example: report1 report2",
+    )
+
+    parser.add_argument(
+        '--sys_names',
```
odashi, line 236: How about following GNU guideline for long options?

> Long options consist of -- followed by a name made of alphanumeric characters and dashes. Option names are typically one to three words long, with **hyphens to separate words**.

https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

explainaboard/visualizers/draw_charts.py
```diff
@@ -0,0 +1,265 @@
...
+    output_dir = "figures" if args.output_dir is None else args.output_dir
+
+    draw_charts_from_reports(reports, output_dir, sys_names=sys_names)
+
+
+if __name__ == '__main__':
+    main()
+    # explainaboard --system_outputs ./data/system_outputs/multilingual/json/
+    # CL-mlpp15out1sum/marc/test-de_9330.json
+    # python draw_charts.py --reports ../log.res
```
odashi, line 265: It looks this is the example of running this script, which should be put on the top of the file.

neubig: Done, I also made it more verbose.

explainaboard/visualizers/draw_charts.py
neubig: This seems reasonable, but currently the whole project is standardized to underbars so we should probably change throughout the project if we want to do this. I opened an issue: https://github.com/neulab/ExplainaBoard/issues/390

explainaboard/visualizers/draw_charts.py
neubig: The reason to apply this restriction is because when creating a confusion matrix it's easier to read one that has the columns and rows in the same order, i.e.:

```
                positive   negative
positive
negative
```

is easier to read than

```
                negative   positive
positive
negative
```

There was no particular reason for using the first element (other than it being easier), so I changed to use the total instead.

docs/visualization.md
```diff
@@ -10,10 +10,11 @@ tools, browsing capabilities, etc. and make analysis relatively easy.
 If you've preferred to use the offline [CLI interface](cli_interface.md), there are
 also some rudimentary visualization tools at your disposal.
 
-**Histograms:** If you want to draw visualizations of the bucketed analysis results
-offline, you can run the following command over one or more reports:
+**Charts:** If you want to draw visualizations of the bucketed analysis results or
+or confusion matrices offline, you can run the following command over one or more
```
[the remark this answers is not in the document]
odashi: @neubig This (repeating "or") should be fixed I guess.

odashi (approved):
Almost OK but there is a remaining comment.

PR #389, opened 2022-08-21
Common implementation of type registry and serializer.

odashi:
This pr attempts to introduce a common implementation of the type registry and the serializer pattern.

The `TypeRegistry` class remembers any `str <-> type` mapping. It also provides a type parameter that restricts the acceptable types (checked by the static type checker).

The `Serializable` interface provides a common way to serialize/deserialize objects.

The `PrimitiveSerializer` convertes nested `Serializable` objects into a data with only builtin types, which can be passed to other string serializer, e.g., `json.dumps()`.

### DIFF
--- a/explainaboard/serialization/registry.py
+++ b/explainaboard/serialization/registry.py
@@ -1,5 +1,7 @@
 """Definition of TypeRegistry."""
 
+from __future__ import annotations
+
 from collections.abc import Callable
 from typing import Generic, TypeVar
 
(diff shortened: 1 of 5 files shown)


neubig (approved):
Overall this looks great, and I'd be OK with merging as-is.

My only comment (which you can reflect if you wish) is that `@registry.register(...)` may be a little bit vague to those who are unfamiliar with the syntax, and you might consider using `@type_registry.register(...)` or `@serialization_registry.register(...)` instead to make the intent of the decorator more clear.

odashi:
@neubig changed `registry` to `test_registry`

neubig (approved):
LGTM

PR #394, opened 2022-08-23
Cache pip dependencies to speed up CI

tetsuok:
This PR attempts to speed up both `unit-tests` and `integration-tests` CI jobs. Every CI job spends about 2 minutes on installing pip packages. The step dominates about 90% of the total time of `unit-tests` and about 30% of the total time of `integration-tests`. The step to install pip packages can be skipped by creating virtual environments and caching the installed packages onto the environments using `actions/cache`. Note that `actions/setup-python@v4` doesn't support caching installed packages. It only allow to avoid re-downloading by caching downloaded packages from PyPI under `~/.cache/pip`.

Dependencies listed in `setup.py` are moved to `requirements.txt`. This is to generate lock files for every Python version from `requirements.txt`. The generated lock files are used as keys to caches to properly invalidate when dependencies are updated. Unless dependencies are changed, every CI job should be reproducible (with respect to installing pip dependencies). Making the CI jobs reproducible and faster achieves at the expense of periodical updates of these lock files. Maintaining lock files for dependencies is pretty common in other programming languages such as JS and Rust. This update can be done by running `cicd/gen_requirements_lock.sh`.


pfliu-nlp:
This looks pretty helpful!

lyuyangh:
This looks very useful! Thanks for implementing it! 

Out of curiosity, do you think it would make sense to use poetry or pipenv (or some other options) to implement this? Or do you think it's overkill? Now that we have a lock file, it might also be helpful to use it for local development and for distribution. I think something like poetry can cover both of those scenarios.

odashi:
@tetsuok Could you add a specific reviewer if you make a readied pull request?

Personally I vote +1 to use poetry over manually implementing the lockfile functionality. Applying poetry we can organize everything in `pyproject.toml` and remove `setup.py`.

tetsuok:
@odashi I can't because I don't have a write access to the repository, as per [documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review).

tetsuok:
@lyuyangh @odashi I agree with using a dedicated tool such as poetry to manage lock files by design. I think that topic needs to be discussed in a separate issue first, because the impact of the change is bigger than caching dependencies in CI. How about the following process?
1. Create a new issue, and discuss migrating to poetry with other members there.
2. Create a PR to switch to poetry and update CI/CD pipelines If we get an agreement. This PR needs to focus on switching to poetry.
3. Update this PR to cache dependencies managed by poetry in GitHub Actions. (This PR must be blocked by until Steps 1 and 2 are finished. I believe the current CI jobs are not intolerably slow to merge this PR as soon as possible)

lyuyangh:
Hi @tetsuok ! That sounds good! I think we can put off this PR for now if you are ok with it.

odashi:
Created issue: #414

@tetsuok I think you basically don't need to make any consensus to make issues because it does never affect the code itself. Feel free to make issues if you realized something.

tetsuok:
@odashi Sure.

tetsuok:
@lyuyangh I'm totally fine.

pfliu-nlp:
hi @tetsuok, thanks for the nice contributions.

Regarding "I can't because I don't have a write access to the repository"
I think you can now. Let me know if you still cannot.

tetsuok:
Closing as we can resolve this issue by a different approach and because this PR is behind main.

PR #395, opened 2022-08-23
Reduce outputs of tests CI step

tetsuok:
Currently, `unittest` step shows lots of output, which prevents developers from quickly identifying test failures if some test fails. This PR reduces the output so that they can easily figure out which test passed and failed. Here's a list of changes:
- Suppress messages with level INFO. The messages are not needed in unit tests because they're not critical by definition
- Suppress progress messages in unit tests which are supposed to run very quickly. If some operations that take in the order of minutes and the operations are tested in unit tests, which indicates we're doing something wrong. If those slow tests are really important, the tests need to be run separately as integration tests, not as unit tests.
- Adds `-v` option to `unittest.main` to show passed tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
           python -m spacy download en_core_web_sm
           python -m nltk.downloader punkt
       - name: unittest
-        run: python -m unittest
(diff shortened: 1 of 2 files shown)


tetsuok:
@neubig @odashi would you take a look?

neubig (approved):
LGTM, thanks a bunch!

odashi (approved):

PR #396, opened 2022-08-23
Rename test files so that unittest finds the files

tetsuok:
Currently, the following tests are not run in the `python -m unittest` command, because the `unittest` only executes the test files matching the pattern, `test*.py` ([reference](https://docs.python.org/3/library/unittest.html#cmdoption-unittest-discover-p)). This PR renames these files so that they are run in the command.

```
explainaboard/loaders/file_loader_test.py
explainaboard/utils/io_utils_test.py
explainaboard/utils/typing_utils_test.py
```

It turned out `TestIOUtils.test_text_writer_file` and `FileLoaderTests` are broken. This PR fixes the tests as well.


lyuyangh:
Thanks for fixing the tests! I didn't know unittest uses "test*.py" by default. 

When I created those test files, I followed this convention:
1. unit tests for a module/file -> name it `{module_name}_test` so it shows up right next to the implementation
2. integration tests -> put it under **/tests** and name it `test_{}`

These three test files fall under the first category. I don't think this convention is documented anywhere for this project but I think it could be useful to distinguish these two kinds of tests. Do you think we should formalize this convention? (This is probably something that we should discuss amongst all the developers, so please feel free to comment on it.)

If we do want to keep this convention, maybe we can update all the `python -m unittest` commands to `python -m unittest -p "*test*.py"` or create a bash script to run all the tests.

odashi:
Hey @lyuyangh @tetsuok , thanks for pointing the problem out.

Although it is better to separate unit tests and integration tests, file name convention needs to be unified to avoid unnecessary context switching. I have no idea either `test*.py` or `*test.py` is better since either have different pros/cons.

> `python -m unittest discover -p "*test*.py"`

This seems the target includes any files containing the string `test` in their names and better not to  be adopted.

odashi:
I also guess we need to include at least 1 underscore to the petterns: `*_test.py` or `test_*.py` to avoid any cases that the word involves `test` as prefix/suffix (e.g., `contest` `testable`).

lyuyangh:
Yeah, that's a good point. In that case, I think we should maybe rename all test files to `xxxx_test.py` so all the unit tests stay right next to the implementation (if the files are sorted alphabetically). And, we would use `*_test.py` as the pattern. What do you think?

odashi:
@lyuyangh SGTM. At this point we need to accept this change to let all the tests work normally, and we would make another pr to fix file names as you suggested.

@tetsuok WDYT?

tetsuok:
@lyuyangh @odashi The scope of this PR is to include the tests that were not run in CI, and fix the broken tests. I think discussing changing the naming conventions and file organizations in this PR need to be discussed separately in Issues first, because they are different topics. Both of @lyuyangh suggestions need the agreement with other maintainers and they need to be documented clearly as coding style and/or contribution guidelines so that contributors can follow the rules. In particular, we need to be careful about changing the naming conventions because developers who are familiar with the default naming pattern would get confused. Unless the coding style/contribution guideline, we probably need to point out the same issues repeatedly in pull requests.

odashi:
Created issue: https://github.com/neulab/ExplainaBoard/issues/398

Would it be great if we merge this change, wouldn't it?

lyuyangh (approved):

odashi (approved):

PR #397, opened 2022-08-23
Add SerializableDataclass

odashi:
This pr adds `SerializableDataclass` mix-in, which provides a common interface to serialize/deserialize dataclasses.

For now `serialize()` and `deserialize()` functions are protected by `@final` to let the derived classes follow the common behavior.

### DIFF
--- a/explainaboard/serialization/test_serializers.py
+++ b/explainaboard/serialization/test_serializers.py
@@ -7,7 +7,11 @@ import unittest
 
 from explainaboard.serialization.registry import TypeRegistry
 from explainaboard.serialization.serializers import PrimitiveSerializer
-from explainaboard.serialization.types import Serializable, SerializableData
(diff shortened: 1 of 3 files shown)


neubig (approved):
Thanks, this looks good!

I'm happy to approve this as-is, but one thing that I considered is that it seems that it's necessary to *both* add the `@dataclasses.dataclass` decorator *and* use `SerializableDataClass`. It seems that developers might accidentally forget to add the decorator and it'd be nice to check to make sure that the decorator is present, perhaps in `SerializableDataClass`'s `serialize` or `deserialize` member functions. I'll leave it up you @odashi about whether this is necessary or not.

odashi:
@neubig sure, it sounds good idea.

neubig (approved):
Great, I approve!

PR #400, opened 2022-08-24
Remoeve additional MetricConfig name

odashi:
This pr removes a behavior in the MetricConfg's registry that registers the name "Foo" as well as "FooConfig". This behavior is unused in the repository, and unnecessarily increases the ambiguity of the config file (if we find the cls_name "Foo", it is unclear whether the class name is really "Foo" or "FooConfig").


neubig (approved):
Good idea, I approve.

PR #402, opened 2022-08-24
Avoid showing the first line of test method's docstring in tests

tetsuok:
Currently, running `test_missing_loader` method shows the following "unusual" output as if there is code to print out messages using `print`:
Screen Shot 2022-08-24 at 22 12 24

Python's unittest module has a behavior that [the test runner prints the first line of test method's docstring in tests](https://github.com/python/cpython/blob/203b598e5113f3662bfbaf3474dd62a258b8b6ce/Lib/unittest/runner.py#L44-L54). It seems better to rename the test name to more informative for now because all tests except this test and `test_mlqa_preprocessor` (this docstring isn't informative) don't use this feature. Whether to use this feature or not can be discussed in separate places if necessary.

### DIFF
--- a/explainaboard/tests/test_loaders.py
+++ b/explainaboard/tests/test_loaders.py
@@ -90,8 +90,9 @@ class BaseLoaderTests(TestCase):
         spaces_pred = loader_pred.load(spaces_path, Source.local_filesystem)
         self.assertEqual(tabs_pred, spaces_pred)
 
-    def test_missing_loader(self):
-        """raises ValueError because a tsv file loader is not provided by default"""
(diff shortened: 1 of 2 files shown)


tetsuok:
Actually, running `test_mlqa_preprocessor` causes the test runner to print out its description. I'll update to remove it.

pfliu-nlp (approved):
Looks good and thank @tetsuok

Regarding >"Whether to use this feature or not can be discussed in separate places if necessary."
Maybe creating an issue first could be considered.

odashi:
Created a related issue about docstring structure itself: https://github.com/neulab/ExplainaBoard/issues/404

Issue #398, opened 2022-08-24
Fixing naming convention of test code files


odashi:
Context:
https://github.com/neulab/ExplainaBoard/pull/396

This issue seems to be also a part of the contribution guideline:
https://github.com/neulab/ExplainaBoard/issues/376

We have currently two options for this convention:

* `test_<name>.<ext>`: `unittest`'s default convention.
* `<name>_test.<ext>`: unit tests have the same prefix name with its target.

odashi:
@lyuyangh @tetsuok @neubig @pfliu-nlp Put comments if any.

odashi:
I don't have strong opinions, but I've basically worked with the second option for a long time.

tetsuok:
I don't have strong opinions about the naming conventions. I'm fine either way as long as some automation can detect filenames that don't follow the naming conventions in CI. For example, if we agree with using the second option, the CI system needs to detect and report files that use the first option.

odashi:
Created #405 to apply several changes.

odashi:
@tetsuok

> the CI system needs to detect and report files that use the first option.

It would be great if we have it, but I guess this also looks too much and it is actually hard to detect appropriately. E.g., imagine that we have `test_utils.py` that provides only utilities for testing but no any test cases.

odashi:
As in #405 we are now using `*_test` format. Let me close this issue.

PR #405, opened 2022-08-24
Restructure test files

odashi:
Related to https://github.com/neulab/ExplainaBoard/issues/398

This PR attempts to change the whole structure of test codes.

- Separate integration tests from `explainaboard` directory. They are now placed on `integration_tests` top-level directory that are processed separately.
- Introduce separate test scripts for unit tests `cicd/run_unit_tests.sh` and integration tests `cicd/run_integration_tests.sh`
- Change test file names from `test_*.py` to `*_test.py`
- Change TestCase class names from `Test*` to `*Test` to follow the same naming convention.

CC: @tetsuok


lyuyangh (approved):
Thanks for making this massive change! I think this also has the added benefit of a smaller package because all the tests and artifacts are excluded when it's distributed.

neubig (approved):
Yes, this looks great to me and I agree with Lyuyang that it'll be good to reduce the distribution package size as well.

pfliu-nlp (approved):
Thanks, and it looks nice!

odashi:
> Thanks for making this massive change! I think this also has the added benefit of a smaller package because all the tests and artifacts are excluded when it's distributed.

Yes this is also intended. To this end we also need to modify `setup.py`. At this moment we'll keep the current script since it is somewhat out of interest of this PR.

PR #413, opened 2022-08-28
Remove py_utils

odashi:
This pull request removes `py_utils.py` and related functions in `summarization.py`. These functions are not used in this repository.


pfliu-nlp (approved):
Looks good, and thanks for the capture. I guess the reason why this code is left here is that I disabled some inefficient features previously but didn't remove the code. It's fine to remove them now since we could retrieve them in history if needed in the future.

odashi:
@pfliu-nlp Great, tahnks for providing the context!

I found this deprecation unexpectedly during other refactoring. It means we need some additional CIs to detect this kind of issues without our burden. Let's take a look at this points later.

PR #418, opened 2022-08-30
Add --skip-failed-analyses option

odashi:
This PR adds ~~`skip_failures`~~ `skip_failed_analyses` flag to `Processor` and ~~`--skip-failures`~~ `--skip-failed-analyses` CLI option to provide a user-side control to choose whether all analyses should succeed or not. Each `Analysis` now returns just `AnalysisResult` or raise errors, and error handling is delegated to `Processor`.

This PR also contains several typing fixes encountered during applying this change.

### DIFF
--- a/explainaboard/analysis/analyses.py
+++ b/explainaboard/analysis/analyses.py
@@ -59,7 +59,7 @@ class Analysis:
         metrics: list[Metric],
         stats: list[MetricStats],
         conf_value: float,
-    ) -> AnalysisResult | None:
+    ) -> AnalysisResult:
         """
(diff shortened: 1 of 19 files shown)


neubig (approved):
The code itself looks good, but perhaps `--skip-failed-analyses` would be more clear as naming, especially as an argument to `explainaboard_main`. I'll leave it up to you whether you think changing to this would be helpful (happy to approve even if it's not changed).

odashi:
> `--skip-failed-analyses`

SGTM. I will change this pr.

pfliu-nlp (approved):
Thanks, this strategy looks good to me.

PR #432, opened 2022-08-31
Delete unused get_metric_list_for_processor

tetsuok:
This PR deletes unused function, `get_metric_list_for_processor`. Leaving unused code prevents us from efficiently refactoring code.

### DIFF
--- a/explainaboard/processors/processor_registry.py
+++ b/explainaboard/processors/processor_registry.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from explainaboard import TaskType
-from explainaboard.metrics.metric import MetricConfig
 from explainaboard.processors.processor import Processor
 
 _processor_registry: dict = {}
@@ -27,8 +26,3 @@ def register_processor(task_type: TaskType):
         return cls
 
     return register_processor_fn
-
-
-def get_metric_list_for_processor(task: TaskType) -> list[MetricConfig]:
-    processor_class = _processor_registry[task]
-    return processor_class.full_metric_list()


pfliu-nlp:
Hi, @tetsuok (cc @neubig) thanks for this PR!
Removing this function will lead to a potential bug in explainaboard_web where the function will be used:
https://github.com/neulab/explainaboard_web/blob/3ff6b6fecc986cb3277eae696558f0803c1300d3/backend/src/impl/default_controllers_impl.py#L17

We can re-consider a good way to do this.

tetsuok:
@pfliu-nlp Oh, I didn't know that `explainaboard_web` directly invokes the function. `get_metric_list_for_processor` is a just shorthand of `get_processor()` and `full_metric_list()`. I think the [invocation](https://github.com/neulab/explainaboard_web/blob/3ff6b6fecc986cb3277eae696558f0803c1300d3/backend/src/impl/default_controllers_impl.py#L93) can be safely replaced with `get_processor(_task.name).full_metric_list()`.

pfliu-nlp (approved):
Since we have this: https://github.com/neulab/explainaboard_web/pull/275
Then it looks good then. Thanks for this PR!

tetsuok:
>  I think, in general, it's OK to have functions that are "unused" in this codebase though because it is an SDK (not just a CLI).

In my experience, leaving unused code caused so much troubles:
- codebase becomes more complex to understand. When someone who knows the context leave the project, it's increasingly difficult to remove unused code.
- Increased complexity leads to productivity drains. It would delay the time to launch new features. Implementing new features sometimes need re-architecture or huge amount of refactoring. Unused code prevents the work. Eventually, some people will leave the project because of the low code health and low productivity, starting everything from scratch.

> There are going to be functions that are "exported" and intended to be used by our users. Maybe we need a better way to identify all the public/exported functions.

Approaches I know are the following:
- [PEP 8](https://peps.python.org/pep-0008/#public-and-internal-interfaces) mentions "internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore."
- List public APIs in the [__init__ module](https://github.com/tetsuok/ExplainaBoard/blob/6a85d3c51fd1f8eca8ddfdcba11d59181a183781/explainaboard/__init__.py), everything else is considered internal/private.

neubig (approved):

Issue #422, opened 2022-08-30
Migrate processor registry to TypeRegistry


tetsuok:
This is a sub-task to migrate processor registry to TypeRegistry.

file: explainaboard/processors/processor_registry.py
symbols: get_processor register_processor get_metric_list_for_processor

List of files to use the symbols:

```
docs/example_scripts/get_customized_feature_results.py:14:16
docs/example_scripts/get_customized_feature_results.py:1:44
docs/example_scripts/test_kg.py:18:12
docs/example_scripts/test_kg.py:3:44
explainaboard/__init__.py:10:5
explainaboard/__init__.py:4:37
explainaboard/explainaboard_main.py:476:24
explainaboard/explainaboard_main.py:9:44
explainaboard/processors/__init__.py:24:56
explainaboard/processors/__init__.py:30:5
explainaboard/processors/aspect_based_sentiment_classification.py:16:56
explainaboard/processors/aspect_based_sentiment_classification.py:20:1
explainaboard/processors/chunking.py:11:1
explainaboard/processors/chunking.py:6:56
explainaboard/processors/cloze_generative.py:24:56
explainaboard/processors/cloze_generative.py:28:1
explainaboard/processors/cloze_multiple_choice.py:23:56
explainaboard/processors/cloze_multiple_choice.py:27:1
explainaboard/processors/conditional_generation.py:36:56
explainaboard/processors/conditional_generation.py:42:1
explainaboard/processors/extractive_qa.py:20:56
explainaboard/processors/extractive_qa.py:24:1
explainaboard/processors/grammatical_error_correction.py:11:56
explainaboard/processors/grammatical_error_correction.py:14:1
explainaboard/processors/kg_link_tail_prediction.py:22:56
explainaboard/processors/kg_link_tail_prediction.py:27:1
explainaboard/processors/language_modeling.py:24:56
explainaboard/processors/language_modeling.py:29:1
explainaboard/processors/machine_translation.py:17:56
explainaboard/processors/machine_translation.py:21:1
explainaboard/processors/named_entity_recognition.py:11:1
explainaboard/processors/named_entity_recognition.py:6:56
explainaboard/processors/nlg_meta_evaluation.py:14:56
explainaboard/processors/nlg_meta_evaluation.py:17:1
explainaboard/processors/processor_registry.py:10:4
explainaboard/processors/processor_registry.py:19:4
explainaboard/processors/processor_registry.py:22:21
explainaboard/processors/processor_registry.py:25:8
[...]

PR #433, opened 2022-08-31
Add test for Processor registry

tetsuok:
This PR adds tests for get_processor in processor registry.  This is a preliminary change of #422.


tetsuok:
Sure. I'll split the test into separate file for each Processor.

odashi (approved):

PR #435, opened 2022-09-01
Split bucket interval

### LINKED ISSUE #431: `bucket_interval` is type-ambiguous
Currently `bucket_interval` is ambiguous in its typing, which is basically `tuple[float,float] | type[str,]` where the former is used for continuous values and latter is used for discrete values. 
https://github.com/neulab/ExplainaBoard/blame/main/explainaboard/analysis/performance.py#L11

We should probably change to the following:
```
bucket_interval: tuple[float,float] | None
bucket_name: str | None
```

odashi:
This pull request splits bucket intervals into two variables: interval ([float, float]) and name (str).

Fixes #431.

### DIFF
--- a/docs/example_scripts/test_kg.py
+++ b/docs/example_scripts/test_kg.py
@@ -57,11 +57,12 @@ for analyses in fine_grained_res:
             confidence_score_high = bucket_performance.confidence_score_high
 
             print("------------------------------------------------------")
+
+            bucket_name = unwrap(bucket_info.bucket_name)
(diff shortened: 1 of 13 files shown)


odashi:
@neubig Could you fix an error around draw_charts.py? It looks a bug but I couldn't guess the correct fix.

neubig:
@odashi done!

neubig (approved):
LGTM, thanks!

odashi:
@neubig Thanks. It looks the change contains a cornercase so I will fix a bit before merging.

PR #436, opened 2022-09-02
Use Serializable for Tokenizers

odashi:
This pull request implements `Serializable` interface to `Tokenizer`s and removes existing code that directly treats dicts.

This PR also fixes some typing/interface issues around `tokenizer.py` encountered during applying this change, and moved code in `integration_tests/tokenizers_test.py` to `explainaboard/utils/tokenizer_test.py` because it is actually unit tests.

### DIFF
--- a/explainaboard/utils/tokenizer.py
+++ b/explainaboard/utils/tokenizer.py
@@ -1,10 +1,15 @@
+"""Definition of tokenizers."""
+
 from __future__ import annotations
 
 import abc
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
 from functools import lru_cache
 import re
 import string
 import sys
+from typing import final, overload
 import unicodedata
 
 from sacrebleu.tokenizers import BaseTokenizer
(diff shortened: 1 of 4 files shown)


explainaboard/utils/tokenizer.py
```diff
@@ -68,55 +120,51 @@ def detokenize(self, tokens: list[str]) -> str:
...
 class SingleSpaceTokenizer(Tokenizer):
     """
     Split a string on a single ascii space
     """
 
     @lru_cache(maxsize=20)
     def __call__(self, text: str) -> TokenSeq:
-        ret = TokenSeq()
         start = 0
+        strs: list[str] = []
```
neubig, line 155: Just out of curiosity, I was wondering why this change is made. Is it because the dataclass was turned into `final`? (no changes to the code are necessary)

neubig (approved):
Looks great, thanks!

explainaboard/utils/tokenizer.py
odashi: It is because of separation of concerns. This function requires no information from Tokenizer because the arguments provide sufficient information to construct TokenSeq.

odashi: Maybe it was better to separate this change from this pr to keep the whole change simple.

neubig: Oh, sorry, my comment was not very clear. I meant: "no changes to this PR are necessary, I'm just asking out of curiosity".

odashi: Nice, thanks for the clarification!

Issue #416, opened 2022-08-29
Text pair classification fails with some DataLab datasets


neubig:
Text pair classification won't work for some DataLab datasets, probably because vocabulary tabulation fails when the name of the first and second textual columns are not `text1` and `text2` (e.g. they may be `premise` and `hypothesis` for NLI). For example, we run the folllowing command:

```
python -m explainaboard.explainaboard_main --task text-pair-classification --system-outputs paws-fra.json --output-file-type json --dataset xtreme --sub-dataset PAWS-X.fra --split test --source-language fra --target-language fra
```

With the attached system output.
[paws-fra.json.zip](https://github.com/neulab/ExplainaBoard/files/9445588/paws-fra.json.zip)

We get the following error:

```
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/gneubig/work/ExplainaBoard/explainaboard/explainaboard_main.py", line 518, in <module>
    main()
  File "/Users/gneubig/work/ExplainaBoard/explainaboard/explainaboard_main.py", line 477, in main
    report = processor.process(
  File "/Users/gneubig/work/ExplainaBoard/explainaboard/processors/processor.py", line 483, in process
    overall_statistics = self.get_overall_statistics(metadata, sys_output)
  File "/Users/gneubig/work/ExplainaBoard/explainaboard/processors/processor.py", line 463, in get_overall_statistics
    external_stats = self._gen_external_stats(sys_info, self._statistics_func)
  File "/Users/gneubig/work/ExplainaBoard/explainaboard/processors/processor.py", line 168, in _gen_external_stats
    new_train = dataset[split_name].apply(  # type: ignore
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/arrow_dataset.py", line 939, in apply
    result = next(self.apply_basic(func))
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/arrow_dataset.py", line 897, in apply_basic
    yield func(self)
[...]

PR #437, opened 2022-09-02
Correct training set feature field names

neubig:
Previously, calculation of training set features would fail if the datalab dataset used unconventional column names.

This does the following things:
1. Makes an option to use `Loader` to load only datasets without system outputs if `output_data` is set to None
2. Changes `_statistics_func` to simply take in the samples and system info, and return the statistics (in contrast to previously using the datalab `aggregating()` functionality.
3. Loads data used in calculating training features through `Loader` so that appropriate field mapping will be performed

Fixes https://github.com/neulab/ExplainaBoard/issues/416

Notably, @pfliu-nlp, "2." may require some discussion, here are the pros and cons of doing it this new way:

**Pros**
* it makes the statistics code self-contained and not rely on an external library. honestly, even though I'm very familiar with explainaboard, I was always a bit confused about what was actually going on here because the `aggregating()` decorator was a bit mysterious to me
* `statistics_func` can now be called on any set of samples, so it could be called on a non-datalab dataset. this may be useful if we want to, for example, calculate training set features with custom datasets

**Cons**
* the datalab aggregating operator may have implemented parallelism so this aggregation of statistics might be able to be done faster? but I actually am not sure if that's actually the case in practice
* something else I'm missing?


neubig:
Sorry, hang on, this implementation was incomplete but I didn't realize it when I first pushed.

pfliu-nlp:
Hi @neubig, I quickly looked through the code, which is good.

>"the datalab aggregating operator may have implemented parallelism so this aggregation of statistics might be able to be done faster?"

it could work [in this way](https://github.com/ExpressAI/DataLab/blob/614aadfac653ed2b6e51c0eeb4d99c5fbf3af8e6/datalabs/tests/arrow_dataset/test_multi.py#L14)
But I think your current implementation looks fine as well since it could solve this issue https://github.com/neulab/ExplainaBoard/issues/416?

explainaboard/analysis/feature_funcs.py
```diff
@@ -74,7 +74,7 @@ def get_lexical_richness(sentence: str):
 
 
 def accumulate_vocab_from_samples(
-    samples: Iterator, text_from_sample: Callable, tokenizer: Tokenizer
+    samples: Iterable, text_from_sample: Callable[..., str], tokenizer: Tokenizer
```
odashi, line 77: Could you add the yield type to `Iterable`?

explainaboard/processors/processor.py
```diff
@@ -110,18 +112,15 @@ def _get_statistics_resources(self, sys_info: SysOutputInfo) -> dict[str, Any]:
         """
         return {"cls": self, "sys_info": sys_info}  #
 
-    @aggregating
-    def _statistics_func(self):
-        return {}
+    @abc.abstractmethod
+    def _statistics_func(self, samples: Iterable, sys_info: SysOutputInfo):
```
odashi, line 116: Could you add the yield type of `Iterable`?

explainaboard/processors/processor.py
```diff
@@ -139,36 +138,25 @@ def _gen_external_stats(
...
+                    loader = get_loader_class(self.task_type()).from_datalab(
+                        DatalabLoaderOption(
+                            sys_info.dataset_name, sub_dataset, split=split_name
+                        ),
+                        output_data=None,
+                    )
+                    dataset = loader.load()
                 except Exception:
-                    dataset = None
+                    pass
```
odashi, line 151: It also looks this block captures any unexpected exceptions. Please narrow the exception type to focus on only expected exceptions, or add comments why this exception can be ignored safely.

odashi:
> because the aggregating() decorator was a bit mysterious to me

It was also unclear to me since the decorator looks to do nothing at a glance of the code.
It is also desirable to avoid dependency on external tools from the core library as well as possible.

> so this aggregation of statistics might be able to be done faster?

It is basically better to avoid any code to improve efficiency if we don't have a strong necessity of it.

neubig:
Thanks for pointing out the over-broad exception @odashi ! I made it less broad and that actually uncovered a few existing bugs 😅 

@pfliu-nlp : there's one bug related to `qa_open_domain` in DataLab (probably not being appropriately added to the task registry). Could you take a look at that?

```
  File "/Users/gneubig/work/ExplainaBoard/integration_tests/qa_open_domain_test.py", line 22, in test_datalab_loader
    data = loader.load()
  File "/Users/gneubig/work/ExplainaBoard/explainaboard/loaders/loader.py", line 154, in load
    dataset_loaded_data = self._dataset_file_loader.load(
  File "/Users/gneubig/work/ExplainaBoard/explainaboard/loaders/file_loader.py", line 335, in load
    raw_data = self.load_raw(data, source)
  File "/Users/gneubig/work/ExplainaBoard/explainaboard/loaders/file_loader.py", line 567, in load_raw
    dataset = load_dataset(
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/load.py", line 2144, in load_dataset
    builder_instance.download_and_prepare(
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/builder.py", line 657, in download_and_prepare
    self.info = self._load_info()
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/builder.py", line 903, in _load_info
    return DatasetInfo.from_directory(self._cache_dir)
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/info.py", line 563, in from_directory
    return cls.from_dict(dataset_info_dict)
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/info.py", line 568, in from_dict
    return cls(**{k: v for k, v in dataset_info_dict.items() if k in field_names})
  File "<string>", line 40, in __init__
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/info.py", line 377, in __post_init__
    templates = [
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/info.py", line 380, in <listcomp>
    else task_template_from_dict(template)
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/tasks/__init__.py", line 85, in task_template_from_dict
    return get_task(task_name).from_dict(task_template_dict)
  File "/Users/gneubig/opt/anaconda3/envs/explainaboard_web/lib/python3.10/site-packages/datalabs/tasks/base.py", line 175, in get_task
    return TASK_REGISTRY[task]
KeyError: 'qa_open_domain'
```

pfliu-nlp:
Hi @neubig have you already fixed this (I saw all the ci tasks have passed) or maybe how to reproduce this?

neubig:
Sorry, it might have been an artifact of my local environment. It seems to be working now.

pfliu-nlp (approved):
looks good to me, and thanks for implementing this idea

odashi (approved):

odashi:
@pfliu-nlp @neubig I also proposed a (maybe) quick fix on cache: https://github.com/ExpressAI/DataLab/pull/340

pfliu-nlp:
@pfliu-nlp thanks approved!

Issue #441, opened 2022-09-03
Bug: the calculation of metric stat for generation tasks


pfliu-nlp:
The calculation of metric stats for generation tasks assumes all metrics are eaas style, which is not true. For example: `ExternalEvalConfig`

https://github.com/neulab/ExplainaBoard/blob/7619c56593f6b14e1a3928e7b433f1db7ff0ba68/explainaboard/processors/conditional_generation.py#L294

pfliu-nlp:
It seems that we have a corresponding implementation: https://github.com/neulab/ExplainaBoard/blob/b9fe82ac8bb710b09ca0d0eaeadfe53f570db688/explainaboard/processors/conditional_generation.py#L327
maybe it's missed due to some merged conflicts.

neubig:
Fixed by https://github.com/neulab/ExplainaBoard/pull/442

PR #442, opened 2022-09-03
Fix the missing metrics of human evaluation

pfliu-nlp:
This PR fixed this issue: https://github.com/neulab/ExplainaBoard/issues/441

### DIFF
--- a/explainaboard/processors/conditional_generation.py
+++ b/explainaboard/processors/conditional_generation.py
@@ -320,7 +320,7 @@ class ConditionalGenerationProcessor(Processor):
                 for i, name in enumerate(metric_names_eaas)
             ]
 
-            # For human metric
+            # For non-EaaS metrics
             for metric_config in metric_configs_noneaas:
                 metric_stats.append(
                     metric_config.to_metric().calc_stats_from_data(
--- a/explainaboard/processors/conditional_generation.py
+++ b/explainaboard/processors/conditional_generation.py
@@ -29,7 +29,7 @@ from explainaboard.metrics.eaas import (
     EaaSMetricStats,
     get_eaas_client,
 )
-from explainaboard.metrics.external_eval import ExternalEvalConfig
+from explainaboard.metrics.external_eval import EXTERNAL_METRICS, ExternalEvalConfig
 from explainaboard.metrics.f1_score import F1ScoreConfig
 from explainaboard.metrics.metric import MetricConfig, MetricStats, SimpleMetricStats
 from explainaboard.processors.processor import Processor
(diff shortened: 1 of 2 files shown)


explainaboard/processors/conditional_generation.py
```diff
@@ -283,26 +300,48 @@ def _gen_cases_and_stats(
...
+                true_data.append(feature_table["reference"])
+                pred_data.append(feature_table["hypothesis"])
+
+            metric_names_automated = []
+            metric_configs_human = []
+            for metric_config in unwrap_generator(analysis_level.metric_configs):
+                if metric_config.name not in EXTERNAL_METRICS:
+                    metric_names_automated.append(metric_config.name)
+                else:
+                    metric_configs_human.append(metric_config)
```
neubig, line 324: I think the actual difference here is probably not between "automated" and "human" but "EaaS" and "not EaaS". What about making the following change, which would also make the code simpler by removing the dependency on `EXTERNAL_METRICS`. Other changes would be need to made in the following parts of the function (probably just the variable names)

```suggestion
            metric_names_eaas = []
            metric_configs_noneaas = []
            for metric_config in unwrap_generator(analysis_level.metric_configs):
                if isinstance(metric_config, EaaSMetricConfig):
                    metric_names_eaas.append(metric_config.name)
                else:
                    metric_configs_noneaas.append(metric_config)
```

explainaboard/processors/conditional_generation.py
```diff
@@ -173,15 +173,32 @@ def default_metrics(
         cls, level='example', source_language=None, target_language=None
     ) -> list[MetricConfig]:
         eaas_defaults = cls._get_default_eaas_strs()
-        defaults: dict[str, list[MetricConfig]] = {
-            'example': [
-                EaaSMetricConfig(
-                    name=x,
-                    source_language=source_language,
-                    target_language=target_language,
+        full_metrics_human = [
```
neubig, line 176: I think these should probably not be included in `default_metrics`, because these won't be calculated by default every time we use ExplainaBoard, right? Rather, they should be added to `full_metric_list()`

neubig (commented):
For future reference (no need to fix it now): this PR has a few extra commits from another branch that have already been added to main. If you could do `git pull origin main` before sending the PR that will make it a bit easier to review on github by removing these extra commits.

explainaboard/processors/conditional_generation.py
pfliu-nlp: good suggestion!

explainaboard/processors/conditional_generation.py
pfliu-nlp: I think you're right!

explainaboard/processors/conditional_generation.py
```diff
@@ -283,26 +286,48 @@ def _gen_cases_and_stats(
...
             )
-            metric_stats: list[MetricStats] = [
+
+            metric_stats: list[Any] = [
                 EaaSMetricStats(name=name, pos=i, eaas_request=async_request)
-                for i, name in enumerate(metric_names)
+                for i, name in enumerate(metric_names_eaas)
             ]
+
+            # For human metric
```
neubig, line 323:
```suggestion
            # For non-EaaS metrics
```

neubig (approved):
Looks good, I approve! With just one small comment fix.

PR #443, opened 2022-09-03
Extractive QA language codes

neubig:
Extractive QA evaluation does some language specific processing, but this:
1. was only using 2-letter language codes, when some datasets like xtreme use 3-letter language codes
2. was throwing an error when the language was not included in one of the supported languages, which seems too strong when there's an obvious backoff strategy of not doing anything language specific

This PR fixes the two above things.

**Important Note:** Previously, the code was assuming that the language was English and doing English-specific processing when the language is set to `None`. However, I don't think this is good behavior. It'd probably be better to do language-agnostic processing when the language is `None` (and this PR changes to this strategy). However, this could be up for discussion.


pfliu-nlp:
>"However, I don't think this is good behavior. It'd probably be better to do language-agnostic processing when the language is None"

This makes sense!

explainaboard/utils/preprocessor.py
```diff
@@ -90,36 +90,31 @@ def process(self, s: str, resources: dict[str, Any]) -> str:
         language = resources['language']
 
         def remove_articles(text, lang):
-            if lang == 'en' or lang is None:
+            if lang in ['en', 'eng']:
```
odashi, line 93: nit: I think relying on a library to normalize the language code is better.
For example: https://pypi.org/project/langcodes/

odashi (approved):

neubig:
@pfliu-nlp : thanks for the comment! just to confirm, do you approve?

PR #445, opened 2022-09-03
Register EaaSMetricConfg to the registry

### LINKED ISSUE #444: Bug: serialization failed for `EaaSMetricConfig`
See following errors
image

#### Potential Reason:

EaaSMetric hasn't been registered as others intendedly, which will result in an error due to: https://github.com/neulab/ExplainaBoard/blob/98d79c976884eee3fbd2a2b1cf76dae4789d9ae3/explainaboard/serialization/serializers.py#L94

hi @odashi I want to know your suggestions about the way to fix this error.

odashi:
This PR revives the registry entry of `EaaSMetricConfig`.

Fixes #444.

### DIFF
--- a/explainaboard/metrics/eaas.py
+++ b/explainaboard/metrics/eaas.py
@@ -10,6 +10,7 @@ import numpy as np
 import sacrebleu
 
 from explainaboard.metrics.metric import Metric, MetricConfig, MetricStats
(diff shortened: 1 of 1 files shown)


pfliu-nlp (approved):
thanks for the quick fixing

PR #450, opened 2022-09-05
Remove is_custom and id members from FeatureType

odashi:
This PR removes `is_custom` and `id` members from `FeatureType` definition. These members seems to be not used in the repository.


neubig (approved):
Yeah, I think these are obsolete/redundant now. Thanks!

I'd like @pfliu-nlp to double-check though.

pfliu-nlp (approved):
looks good and thanks!

PR #451, opened 2022-09-05
Bump verison to 0.11.1

OscarWang114:
Bumping the version to allow us to start working on updating the web interface by sending PRs.


neubig:
Could we make this 0.11.1? It's not clear that this is a release that has a coherent set of functionality changes, which is kinda what we've been using for the middle version number.

OscarWang114:
Ah sure, I was following semantic versioning and regarded "Add --skip-failed-analyses option" (https://github.com/neulab/ExplainaBoard/pull/418) as new functionality, but it makes sense to follow what we've been using!

neubig (approved):

PR #452, opened 2022-09-06
Add the evaluation for text-to-sql task (spider dataset)

shuaichenchang:
Contains 2 evaluation metrics for SQL prediction:
1) Exact Set Match Accuracy (EM)
2) Execution Accuracy (EX)
The definition can be found in https://arxiv.org/pdf/1809.08887.pdf 
The evaluation scripts are adapted from https://github.com/taoyds/test-suite-sql-eval


pfliu-nlp:
Overall, it looks very nice!! thanks for your contribution!
We can have a meeting to discuss more details besides the above general comments.
Also, the file `question.json` is too large, I will put it in S3 so that we can download and cache it first.

explainaboard/metrics/auxiliary/sql_evaluation/exec_eval.py
```diff
@@ -0,0 +1,268 @@
+import asyncio
```
odashi, line 1: It looks these files are borrowed from somewhere we don't recognize, but there are no explanation about borrowing. Please apply following changes:

- Move `explainaboard/metrics/auxiliary` to `explainaboard/contrib` to indicate this is a third-party.
- Check if the original license is compatible with the MIT license. If so, add the original licensing notes to this folder. If not, we can not borrow the code from such source.

explainaboard/metrics/sql_em_ex.py
```diff
@@ -0,0 +1,69 @@
...
+    table_path: str = ""
+    etype: str = "exec"
+
+    def to_metric(self):
+        return SQLEx(self)
+
+
+class SQLEx(Metric):
+    """
+    Calculate exact set match (Em) and execuation (Ex) accuracy,
```
odashi, line 59:
```suggestion
    Calculate exact set match (Em) and execution (Ex) accuracy,
```

explainaboard/metrics/sql_em_ex_test.py
```diff
@@ -0,0 +1,17 @@
+"""Tests for explainaboard.metrics.sql_em_ex"""
```
odashi, line 1: Could you add unit tests for `calc_stats_from_data`?

explainaboard/processors/text_to_sql.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        return data_point["true_sql"]
+
+    def _get_predicted_label(self, data_point):
+        """
+        Get the predicted label from a data point. Overloaded from parent class.
+        :param data_point: the data point under consideration
+        :return: the predicted label for the output
+        """
+        return data_point["predicted_sql"]
```
odashi: These functions are private and not used in this class.

```suggestion
```

integration_tests/metric_test.py
```diff
@@ -260,3 +261,49 @@ def test_qa_metrics(self):
             2,
             "almost equal",
         )
+
+    @unittest.skip(reason="dependency libs are not installed")
```
odashi, line 265: Could we re-enable all skipped tests? In general, invalidated tests are not very useful to check consistency of the software.

If we really need to disable these tests, please add following things:

- which library is required to invoke this test
- why the library is not added to the dependency

integration_tests/metric_test.py
```diff
@@ -260,3 +261,49 @@ def test_qa_metrics(self):
...
+    def test_sql_em(self):
+        # python -m unittest integration_tests.metric_test.MetricTest.test_sql_em
+        metric = explainaboard.metrics.sql_em_ex.SQLEmConfig(
+            name="SQLEm",
+            db_dir="https://expressai-xlab.s3.amazonaws.com/large_data/database",
```
odashi, line 270: This is considered as a personal storage. We also have a Cloud Storage to store large data related to our software. @pfliu-nlp and @neubig have write access to it.

pfliu-nlp:
@odashi This PR is still modified.  @shuaichenchang could you make this PR  a draft version?

PR #456, opened 2022-09-06
Add task argument pair identification

pfliu-nlp:
#### Please Review this PR first: https://github.com/neulab/ExplainaBoard/pull/454

This is a simple task that can re-use existing metrics (accuracy)

* task definition: Given an argument, the task aims to identify one matched argument from a list of arguments.
* evaluation metric: Accuracy
* dataset: [iapi](https://github.com/ExpressAI/DataLab/blob/main/datasets/iapi/iapi.py)


neubig (commented):
Just a note that I'll review this once the discussion in #454 is merged. Please re-request review then!

neubig:
@pfliu-nlp : could you do `git pull origin main` to merge in the changes from #454 and then re-request review on this PR?

explainaboard/processors/ranking.py
```diff
@@ -0,0 +1,140 @@
...
+                vocab[w] = vocab.get(w, 0.0) + 1.0
+
+            total_samps += 1
+
+        # the rank of each word based on its frequency
+        sorted_dict = {
+            key: rank
+            for rank, key in enumerate(sorted(set(vocab.values()), reverse=True), 1)
+        }
+        vocab_rank = {k: sorted_dict[v] for k, v in vocab.items()}
```
odashi, line 135: This is a bit hard to read. The suggestion is for just improving readability:
```suggestion
        vocab_uniq_counts = sorted(set(vocab.values()), reverse=True)
        vocab_count_ranks = {
            count: rank for rank, count in enumerate(vocab_uniq_counts, 1)
        }
        vocab_rank = {k: vocab_count_ranks[count] for k, count in vocab.items()}
```

explainaboard/processors/ranking.py
```diff
@@ -0,0 +1,140 @@
...
+
+        for k, v in length_fre.items():
+            length_fre[k] = v * 1.0 / total_samps
```
odashi, line 138: - It isn't good to reuse the same variable for different meanings.
- Single slash represents division of floats. We basically don't need the magic constant `1.0`.
```suggestion
        # length_fre should be renamed to length_counts or something similar.
        length_freq = {k: v / total_samps for k, v in length_fre.items()}
```

odashi: Maybe the L138 runs into zero-division error. Some workaround should be implemented.

explainaboard/metrics/ranking.py
```diff
@@ -38,7 +38,7 @@ def calc_stats_from_rank(
 
 
 @dataclass
-@metric_config_registry.register("HitsConfig")
+@metric_config_registry.register("Hits")
```
neubig, line 41: Per our previous discussion, should this be HitsConfig? Similarly, below.

neubig:
Did you see my comment above about "Ranking"?

pfliu-nlp:
Do you mean renaming `RankingProcessor` to `ArgumentPairIdentificationProcessor`? That's something being reflected in the new commit.

neubig:
Ah, I see. It seems that there is still a "RankingLoader" as well. I was suggesting all of the stuff marked as "ranking" could be removed.

pfliu-nlp:
I keep `RankingLoader` unchanged designedly, which could be by two specific tasks: https://github.com/ExpressAI/DataLab/blob/24cf5318ff089691c3fa93c9568aee125e6b967d/datalabs/tasks/ranking.py#L10

explainaboard/loaders/ranking.py
```diff
@@ -0,0 +1,46 @@
...
+
+@register_loader(TaskType.argument_pair_identification)
+@register_loader(TaskType.ranking)
+class RankingLoader(Loader):
+    """Loaders for the ranking tasks, such as argument pair identification."""
+
+    @classmethod
+    def default_dataset_file_loaders(cls) -> dict[FileType, FileLoader]:
+        """See Loader.default_dataset_file_type."""
+        target_field_names = ["context", "utterance", "true_label"]
```
neubig, line 26: Are these `target_field_names` applicable to other ranking tasks? For example, what about the most typical ranking task of information retrieval?

There I feel that in this case we have only `query` and `true_label`?

Maybe to future-proof the naming, it would be more appropriate to name it `ranking_with_context` and `utterance` should be replaced with `query`?

Also apologies for all the iteration on this, but if we set this now and change it later it might affect the validity of the saved DB entries, so we probably want to try to get it right the first time.

neubig (commented):
Overall looks very nice, but I have one comment on the `RankingLoader`!

explainaboard/processors/processor_factory.py
```diff
@@ -66,6 +69,8 @@
     TaskType.tabular_classification: TabularClassificationProcessor,
     TaskType.argument_pair_extraction: ArgumentPairExtractionProcessor,
     TaskType.meta_evaluation_nlg: MetaEvaluationNLGProcessor,
+    TaskType.argument_pair_identification: ArgumentPairIdentificationProcessor,
+    TaskType.ranking_with_context: ArgumentPairIdentificationProcessor,
```
neubig, line 73: This line seems incorrect? Because there's no general purpose `ranking_with_context` processor yet it'd probably be better to just delete this line and allow an error to be thrown if someone tries to specify this task.

```suggestion
```

neubig (approved):
Overall looks great, just one tiny suggestion!

explainaboard/processors/processor_factory.py
pfliu-nlp: thanks a lot for fixing the docs lint issue I also fix this comment!

PR #457, opened 2022-09-06
Add notes for machine translation in README

noelchen90:
This PR adds a note for pair-wise analysis use case for machine translation.
Specifically, we need to bring up that the number of `custom-dataset-paths` need to equal to `system-outputs`.


pfliu-nlp (approved):
Looks good, and thanks for detailing this!

PR #459, opened 2022-09-09
add documentation for custom features

qjiang002:
Add doc and example json file for add custom features and analysis

### DIFF
--- a/data/system_outputs/text-classification-custom-feature-example.json
+++ b/data/system_outputs/text-classification-custom-feature-example.json
@@ -3,18 +3,22 @@
     "custom_features": {
       "example": {
         "verb": {
+          "cls_name": "Value",
           "dtype": "string",
           "description": "the verb in the example"
         },
         "subject": {
+          "cls_name": "Value",
           "dtype": "string",
           "description": "the subject in the example"
         },
         "adverb": {
+          "cls_name": "Value",
           "dtype": "string",
           "description": "the adverb in the example"
         },
         "LL_delta": {
+          "cls_name": "Value",
           "dtype": "float",
           "description": "log likelihood / confidence"
         }
--- a/data/system_outputs/text-classification-custom-feature-example.json
+++ b/data/system_outputs/text-classification-custom-feature-example.json
@@ -1,0 +1,1029 @@
+{
+  "metadata": {
+    "custom_features": {
+      "example": {
+        "verb": {
+          "dtype": "string",
+          "description": "the verb in the example"
+        },
+        "subject": {
+          "dtype": "string",
+          "description": "the subject in the example"
+        },
+        "adverb": {
+          "dtype": "string",
+          "description": "the adverb in the example"
+        },
+        "LL_delta": {
+          "dtype": "float",
+          "description": "log likelihood / confidence"
+        }
+      }
+    },
+    "custom_analyses": [
+      {
+        "cls_name": "BucketAnalysis",
+        "level": "example",
+        "feature": "verb",
+        "num_buckets": 15,
+        "method": "discrete"
+      },
+      {
+        "cls_name": "BucketAnalysis",
+        "level": "example",
+        "feature": "subject",
+        "num_buckets": 50,
+        "method": "discrete"
+      },
+      {
+        "cls_name": "BucketAnalysis",
+        "level": "example",
+        "feature": "adverb",
+        "num_buckets": 10,
+        "method": "discrete"
+      },
+      {
+        "cls_name": "BucketAnalysis",
+        "level": "example",
+        "feature": "LL_delta",
+        "num_buckets": 3,
+        "method": "continuous"
+      }
+    ]
+  },
+  "examples": [
+    {
+      "predicted_label": "patient",
+      "verb": "read",
+      "subject": "orator",
+      "adverb": "nicely",
+      "LL_delta": -1.506662368774414
+    },
+    {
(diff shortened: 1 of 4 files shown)


docs/add_custom_features.md
```diff
@@ -0,0 +1,66 @@
+# Add custom features for custom analysis
+
+If you want to perform custom analysis with your custom features that are not supported in the original task processors, and these features are only related to one particular system/dataset, you can define the custom features and analysis in the `metadata` section in the output JSON file. 
+
+Here is the template of the output JSON file with custom metadata.
```
neubig, line 5: This is nice, but it might be even more clear if you separately add one example of a discrete feature and one example of a continuous feature (and explicitly explain that you're doing so in the text).

docs/add_custom_features.md
```diff
@@ -0,0 +1,66 @@
...
+      "output-json-key": "model-output",
+      "custom-feature-name": "feature-value"
+    }
+  ]
+}
+```
+where
+* `feature-level` represents the fine-grained level of the analysis
+    * `example` for sentence-level analysis
+    * `span` for span-level analysis
```
neubig, line 38:
```suggestion
    * `span` for span-level analysis (e.g. in named entity recognition analysis)
    * `token` for token-level analysis (e.g. in named entity conditional text generation)
```

docs/add_custom_features.md
```diff
@@ -0,0 +1,66 @@
...
+* `dtype` represents the data type of the feature
+    * `float` for continuous feature
+    * `string` for discrete feature
+* `cls_name` means the analysis type
+    * `BucketAnalysis` requires the following fields to be defined
+    ```
+    BucketAnalysis(
```
neubig, line 45: It might be a little bit confusing to suddenly have Python code here, as the expected input format is JSON. I'd suggest just having a bullet list. Similarly below.

neubig (commented):
Thanks a lot, this is great! I have a few comments/suggestions.

qjiang002:
In the second commit:
- Refine doc
- For ComboCountAnalysis, the json file only takes `list` but in the code it takes `tuple`. Add a type casting in analysis.py pass features as a tuple.

docs/add_custom_features.md
```diff
@@ -0,0 +1,125 @@
+# Add custom features for custom analysis
+
+If you want to perform custom analysis with your custom features that are not supported in the original task processors, and these features are only related to one particular system/dataset instead of the task itself, you can define the custom features and analysis in the `metadata` section in the output JSON file. 
+
+## Example of bucket analysis with discrete feature
```
neubig, line 5: Sorry to suggest changes multiple times, but could we give an example that has discrete features and continuous features in the same section/JSON? It seems that splitting it into two sections is a bit repetitive.

Otherwise this looks great, and I'm happy to approve after this is resolved!

data/system_outputs/text-classification-custom-feature-example.json
```diff
@@ -0,0 +1,1035 @@
...
+          "description": "the subject in the example"
+        },
+        "adverb": {
+          "dtype": "string",
+          "description": "the adverb in the example"
+        },
+        "LL_delta": {
+          "dtype": "float",
+          "description": "log likelihood / confidence"
+        }
```
odashi, line 20: The newest version of the repository requires to always specify "cls_name" for every feature.
```suggestion
        "verb": {
          "cls_name": "Value",
          "dtype": "string",
          "description": "the verb in the example"
        },
        "subject": {
          "cls_name": "Value",
          "dtype": "string",
          "description": "the subject in the example"
        },
        "adverb": {
          "cls_name": "Value",
          "dtype": "string",
          "description": "the adverb in the example"
        },
        "LL_delta": {
          "cls_name": "Value",
          "dtype": "float",
          "description": "log likelihood / confidence"
        }
```

docs/add_custom_features.md
```diff
@@ -0,0 +1,125 @@
...
+Here is the output json format for bucket analysis with custom discrete features with `"dtype": "string"`, such as the subject of the sentence, etc.
+```
+{
+  "metadata": {
+    "custom_features": {
+      "feature-level": {
+        "custom-feature-name": {
+          "dtype": "string",
```
odashi, line 13: ditto

docs/add_custom_features.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Example of bucket analysis with continuous feature
+Here is the output json format for bucket analysis with custom continuous features with `"dtype": "float"`, such as the count of particular words, the output logits/probability, etc.
+```
+{
+  "metadata": {
+    "custom_features": {
+      "feature-level": {
+        "custom-feature-name": {
+          "dtype": "float",
```
odashi, line 53: ditto

data/system_outputs/text-classification-custom-feature-example.json
qjiang002: Thanks for the correction. What does "cls_name" mean for each feature? Is it the same as feature name or is it always 'Value'?

odashi: "cls_name" is used to distinguish which type we have to map these objects to. If the feature is either number or string, it should be "Value".

pfliu-nlp (approved):
Thanks for the revision!

odashi (approved):

neubig (approved):

PR #460, opened 2022-09-09
Create CODEOWNERS

odashi:
This PR adds CODEOWNERS that specifies who will be required to review.

@neubig and @pfliu-nlp are specified as global owners.
@odashi is specified as a code maintainer.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,5 @@
 # Global owner
 * @neubig @pfliu-nlp
 
-# Code quality maintainer
+# Code maintainer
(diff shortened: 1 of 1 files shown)


neubig (approved):
This looks great!

Just to clarify the procedure, if other people become familiar with individual pieces of the code, they will be added as codeowners for other portions of the repo, right?

odashi:
@neubig Almost yes, but they should be added only when they can take full responsibility of the portion.

neubig:
Right, makes sense.

PR #469, opened 2022-09-10
refactor AnalysisResult.print to generate_report.

odashi:
This PR replaces `AnalysisResult.print()` with `AnalysisResult.generate_report()` that generates multi-line string of corresponding reports.

Applying this change, `AnalysisResult` no longer has any direct side-effects to the environment, and the report string becomes testable as this PR added some unit tests. 

This PR also fixes some typing ambiguities around `AnalysisResult`s.


neubig (approved):
Good idea, thanks!

PR #470, opened 2022-09-10
Move external_stats from MetricConfig to ExternalEvalConfig

tetsuok:
As per the discussion in #428, this PR moves `external_stats` property from the `MetricConfig` class to the `ExternalEvalConfig` class. This is for separation of concerns: Derived classes from `MetricConfig` except `ExternalEvalConfig` don't need to know the property because they don't use that property. ~This PR also changes the type of the property from `numpy.ndarray` to `list[list[int]]` to make `ExternalEvalConfig` serializable.~

### DIFF
--- a/explainaboard/metrics/external_eval.py
+++ b/explainaboard/metrics/external_eval.py
@@ -39,9 +39,7 @@ class ExternalEvalConfig(MetricConfig):
     instruction: str = "Annotation instruction"
     # The external statistics for metrics. The row size is equal to the number
     # of test samples, the column size is equal to `n_annotators`.
-    # NOTE: Use of `list` rather than `numpy.ndarray` is to make this class
-    # serializable.
-    external_stats: list[list[int]] | None = None
+    external_stats: np.ndarray | None = None
 
     def to_metric(self):
         return ExternalEval(self)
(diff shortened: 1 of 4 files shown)


explainaboard/metrics/external_eval.py
```diff
@@ -37,6 +37,11 @@ class ExternalEvalConfig(MetricConfig):
     n_annotators: int = 3
     categories: int = 5
     instruction: str = "Annotation instruction"
+    # The external statistics for metrics. The row size is equal to the number
+    # of test samples, the column size is equal to `n_annotators`.
+    # NOTE: Use of `list` rather than `numpy.ndarray` is to make this class
+    # serializable.
+    external_stats: list[list[int]] | None = None
```
odashi, line 44: Add `__post_init__` function to validate these conditions.
I'm also worried that maybe double-list is not efficient compared with ndarray in terms of shape checking: ndarray guarantees that every subarray has the same remaining shape, while nested lists does not.

odashi:
@tetsuok ndarrays can be serialized by manually converting it to builtin lists.

```python
class Something(Serializable):
    _myarray: np.ndarray

    def __init__(self, *, myarray: np.ndarray):
        self._myarray = myarray

    def serialize(self):
        return {"myarray": self._myarray.tolist()}

    @classmethod
    def deserialize(cls, data):
        return cls(myarray=np.array(data["myarray"]))

tetsuok:
@odashi Thanks! I reverted to use `numpy.ndarray`. I'll update #428 to serialize/deserialize `ExternalEvalConfig` as you suggested.

odashi (approved):

PR #474, opened 2022-09-11
Remove unused EXTERNAL_METRICS

tetsuok:
`EXTERNAL_METRICS` was introduced in #355, but the variable is no longer used in the codebase. This PR removes the unused list.


odashi (approved):
I didn't know the original intent of this value, but I approve this change so long as all tests are passed.

odashi:
@pfliu-nlp 's approval is required since he is the original author.

pfliu-nlp (approved):
It probably will be used for explainaboad later but let's delete it currently.

PR #475, opened 2022-09-11
Disambiguate "conf"

odashi:
This pull request attempts to remove ambiguation around "conf" in the repository. This abbreviation is used for different meanings: confidence, confusion, and config. Using full name is preferred to avoid user-side confusion.

This pull request also changes the name of arguments `conf_value` to `confidence_alpha`, and fix descriptions to avoid "p-value" wording. In principle, the confidence level (or confidence coefficient) is a constant $(1-\alpha)$ defined before any analysis, while p-value is a result produced by some statistical test. They are pointing to different concepts.

### DIFF
--- a/data/reports/report_kg.json
+++ b/data/reports/report_kg.json
@@ -8,7 +8,7 @@
   "target_language": null,
   "reload_stat": true,
   "is_print_case": true,
-  "conf_value": 0.05,
+  "confidence_level": 0.05,
   "system_details": null,
   "metric_configs": [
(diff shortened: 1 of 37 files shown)


odashi:
I thought `confidence_alpha` is much better than `confidence_level` because the latter makes another confusion.

pfliu-nlp (approved):
Looks good. We probably need a markdown file to maintain the mapping between terminology<->python variables.

PR #476, opened 2022-09-11
Bucket analysis: Handle corner case where samples list is empty

OscarWang114:
Required by [this PR](https://github.com/neulab/explainaboard_web/pull/318) in web.

When a user provides custom bucket intervals, the interval range may be too small that there are no samples within it. This causes `bucket_collection.samples` to become empty.

The previous code did not handle this case and will throw an error in `metric_stat.filter(bucket_collection.samples)`, where `data[indices_array.flatten()]` complains "arrays used as indices must be of integer (or boolean) type" because `indices_array` is empty (cannot perform slicing using an empty array).

I thought it'd be better to handle the empty corner case at a higher level and added the logic in `perform()`.

(Initially forgot to install pre-commit locally. It's installed now)


explainaboard/analysis/analyses.py
```diff
@@ -212,21 +213,30 @@ def perform(
...
-                metric_result = metric_func.evaluate_from_stats(
-                    bucket_stats,
-                    conf_value=conf_value,
-                )
+                # Samples may be empty when user defined a bucket interval that
+                # has no samples
+                if n_samples == 0.0:
+                    value = 0.0
+                    conf_low: Optional[float] = 0.0
+                    conf_high: Optional[float] = 0.0
```
odashi, line 221: It looks setting these values to `None` is better, because any values have no meaning under empty buckets. the low/high bound can become None, but maybe we need to change the type of value to `float | None`.

OscarWang114: Thanks! I initialized the `conf`s to None.

I agree with initializing value to None as well, but it may involve more changes in both SDK and the web interface. I'll open an issue and create a separate PR for it. Does this sound good to you?

OscarWang114:
Hey @odashi do you think this PR could be merged first (needs at least one approval)? As you mentioned, #482 is resolvable by a schema update in `MetricResult`.

odashi (approved):
Thanks!

Issue #465, opened 2022-09-09
Document how custom dataset should be formatted


pfliu-nlp:
Currently, we don't describe how custom datasets (for example: [this](https://github.com/neulab/ExplainaBoard/blob/main/docs/task_argument_pair_extraction.md)) should be organized in the doc that describes each task's analysis setup.

PR #481, opened 2022-09-12
add more concrete documents for dataset formats when performing analyses

pfliu-nlp:
Currently, we don't describe how custom datasets should be organized in the doc that describes each task's analysis setup. This prevents new users from using the custom datasets.

Relevant issues: 
* https://github.com/neulab/ExplainaBoard/issues/465
* https://github.com/neulab/explainaboard_web/issues/300

### DIFF
--- a/docs/concepts_about_system_analysis.md
+++ b/docs/concepts_about_system_analysis.md
@@ -1,0 +1,53 @@
+# Concepts about System Analysis
+In this docs, we will explain some important concepts that
+you need to know before performing system analyses using Explainaboard.
+
+## What a "dataset" and "system output" is
+To perform an analysis of your results, usually, two types of files should be prepared: `dataset`
+and `system output` 
+
+### `Dataset`
+`Dataset` usually consists of test samples together with true labels (or references in text generation
+tasks).
+For example, in text classification, the `dataset` could be organized in a tsv format:
+```text
+I love this movie   positive
+The movie is too long   negative
+...
+``` 
+
+### `System output`
+`System output` is frequently composed of predicted labels (or hypotheses, e.g., system-generated text),
+but sometimes `system output` will also contain test samples, such as `CoNLL` format in sequence labeling tasks.
+For example, in text classification, the `system output` could be organized in a text format:
+```text
+positive
+negative
+...
+```
+
+
+
+## How the dataset can be gotten from two sources, either "datalab" or "custom"
+The `dataset` typically could be obtained from two sources (so far): one is supported by [Datalab](https://aclanthology.org/2022.acl-demo.18.pdf) (Xiao et al 2022) and the
+other is custom.
+* if your datasets have been supported by [datalab](https://github.com/ExpressAI/DataLab/tree/main/datasets),
+you, fortunately, don't need to prepare the dataset. Instead, you just need to get the dataset name for later use.
+For example, the following command can be used to perform system analysis on `sst2` dataset (supported by datalab)
+```shell script
+explainaboard --task text-classification --dataset sst2 --system-outputs ./data/system_outputs/sst2/sst2-lstm-output.txt
+```
+If the dataset hasn't been supported by datalab, we then need to specify a file path:
+```shell script
+explainaboard --task text-classification --custom-dataset-paths ./data/system_outputs/sst2/sst2-dataset.tsv --system-outputs ./data/system_outputs/sst2/sst2-lstm-output.txt
+```
+
+* if your datasets haven't been supported by datalab but you want them supported, you can follow this 
+[doc](https://github.com/ExpressAI/DataLab/blob/main/docs/SDK/add_new_datasets_into_sdk.md) to add them.
+
+
+
+
+## How different formats (e.g. json, tsv, etc.) are supported
+In ExplainaBoard, users are allowed to prepare custom datasets with different formats, such as
+`json`, `tsv` etc. We will detail this in each task's description.
--- a/docs/concepts_about_system_analysis.md
+++ b/docs/concepts_about_system_analysis.md
@@ -1,15 +1,15 @@
 # Concepts about System Analysis
-In this docs, we will explain some important concepts that
+In these docs, we will explain some important concepts that
 you need to know before performing system analyses using Explainaboard.
 
-## What a "dataset" and "system output" is
-To perform an analysis of your results, usually, two types of files should be prepared: `dataset`
-and `system output` 
+## What are a "dataset" and a "system output"?
+To perform an analysis of your results, usually, two types of files should be prepared: "dataset"
+and "system output" 
 
 ### `Dataset`
-`Dataset` usually consists of test samples together with true labels (or references in text generation
-tasks).
-For example, in text classification, the `dataset` could be organized in a tsv format:
+A "dataset" usually consists of test inputs together with true outputs (e.g. gold-standard labels for classification tasks 
+or reference outputs for text generation tasks).
+For example, in text classification, a `dataset` organized in tsv format may look like this:
 ```text
 I love this movie   positive
 The movie is too long   negative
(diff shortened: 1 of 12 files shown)


neubig:
I think this is really nice, but what do you think about having a general purpose explanation of

1. what a "dataset" and "system output" is
2. how the dataset can be gotten from two sources, either "datalab" or "custom"
3. how different formats (e.g. json, tsv, etc.) are supported

Then on each task page, we could link to this explanation and ask people to read it first before going into task-specific particulars.

The thing I'm a little bit worried about with the current format is that it seems the same content is copy-pasted over many different doc files, which means that if we have to revise the content we have to revise every single file.

pfliu-nlp:
Sounds good. I will revise them.

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
+# Concepts about System Analysis
+In this docs, we will explain some important concepts that
```
neubig, line 2:
```suggestion
In these docs, we will explain some important concepts that
```

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+you need to know before performing system analyses using Explainaboard.
+
+## What a "dataset" and "system output" is
```
neubig, line 5:
```suggestion
## What are a "dataset" and a "system output"?
```

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+To perform an analysis of your results, usually, two types of files should be prepared: `dataset`
```
neubig, line 6:
```suggestion
To perform an analysis of your results, usually, two types of files should be prepared: "dataset"
```

I think that `dataset` should mainly be used when "dataset" indicates a piece of code or a variable, which is not the case here. Probably just double quotes is fine? Similarly below.

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+and `system output` 
+
+### `Dataset`
+`Dataset` usually consists of test samples together with true labels (or references in text generation
+tasks).
```
neubig, line 11:
```suggestion
A "dataset" usually consists of test inputs together with true outputs (e.g. gold-standard labels for classification tasks or reference outputs for text generation
tasks).
```

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+For example, in text classification, the `dataset` could be organized in a tsv format:
```
neubig, line 12:
```suggestion
For example, in text classification, a `dataset` organized in tsv format may look like this:
```

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+```text
+I love this movie   positive
+The movie is too long   negative
+...
+``` 
+
+### `System output`
+`System output` is frequently composed of predicted labels (or hypotheses, e.g., system-generated text),
```
neubig, line 20:
```suggestion
`System output` is usually composed of predicted labels (or hypotheses, e.g., system-generated text),
```

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+but sometimes `system output` will also contain test samples, such as `CoNLL` format in sequence labeling tasks.
```
neubig, line 21:
```suggestion
but sometimes `system output` will also contain inputs, such as `CoNLL` format in sequence labeling tasks.
```

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+For example, in text classification, the `system output` could be organized in a text format:
+```text
+positive
+negative
+...
+```
+
+
+
+## How the dataset can be gotten from two sources, either "datalab" or "custom"
```
neubig, line 31:
```suggestion
## "Datalab datasets" and "custom datasets"
```

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+The `dataset` typically could be obtained from two sources (so far): one is supported by [Datalab](https://aclanthology.org/2022.acl-demo.18.pdf) (Xiao et al 2022) and the
```
neubig, line 32:
```suggestion
ExplainaBoard currently supports two sources for datasets: one is through [Datalab](https://aclanthology.org/2022.acl-demo.18.pdf) (Xiao et al 2022) and the
```

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+other is custom.
+* if your datasets have been supported by [datalab](https://github.com/ExpressAI/DataLab/tree/main/datasets),
+you, fortunately, don't need to prepare the dataset. Instead, you just need to get the dataset name for later use.
+For example, the following command can be used to perform system analysis on `sst2` dataset (supported by datalab)
+```shell script
+explainaboard --task text-classification --dataset sst2 --system-outputs ./data/system_outputs/sst2/sst2-lstm-output.txt
+```
+If the dataset hasn't been supported by datalab, we then need to specify a file path:
```
neubig, line 40:
```suggestion
If the dataset hasn't been supported by datalab, we then need to prepare the dataset in a format supported by ExplainaBoard (this varies from task-to-task, see the task-specific documentation) and specify a file path:
```

docs/task_aspect_based_sentiment_classification.md
```diff
@@ -1,13 +1,44 @@
...
+
+* (2) `json` (basically, it's a list of dictionaries with three keys: `aspect`, `text` and `true_label`)
+```json
+[
+  {"aspect":"Boot time", "text": "Boot time  is super fast, around anywhere from 35 seconds to 1 minute.", "true_label": "positive"},
+  ...
+]
+```
+
+* (3) `datalab`: if your datasets have been supported by [datalab](https://github.com/ExpressAI/DataLab/tree/main/datasets),
```
neubig, line 35: I would suggest that we put "datalab" first every time, as this is the preferred format to use if it's supported.

docs/concepts_about_system_analysis.md
```diff
@@ -0,0 +1,53 @@
...
+...
+```
+
+
+
+## How the dataset can be gotten from two sources, either "datalab" or "custom"
+The `dataset` typically could be obtained from two sources (so far): one is supported by [Datalab](https://aclanthology.org/2022.acl-demo.18.pdf) (Xiao et al 2022) and the
+other is custom.
+* if your datasets have been supported by [datalab](https://github.com/ExpressAI/DataLab/tree/main/datasets),
+you, fortunately, don't need to prepare the dataset. Instead, you just need to get the dataset name for later use.
```
neubig, line 35:
```suggestion
* if your datasets have been supported by [datalab](https://github.com/ExpressAI/DataLab/tree/main/datasets),
we recommend that you load it through DataLab. This is for several reasons: (1) you don't have to prepare the data yourself, (2) you can be sure that your accuracy numbers will be comparable with those calculated by other people using ExplainaBoard, and (3) DataLab datasets support analysis with features calculated over the training set, which can be informative.
```

neubig (approved):
Looks great! I did some suggestions and proofreading, but feel free to merge if you're happy with this.

docs/concepts_about_system_analysis.md
pfliu-nlp: excellent!

docs/task_aspect_based_sentiment_classification.md
pfliu-nlp: agree and done

PR #483, opened 2022-09-13
enable Chinese tokenizer to work

pfliu-nlp:
Currently, the tokenizer for Chinese on other tasks (such as aspect-based sentiment classification) hasn't been turned on.
This is just one way to fix this issue, feel free to make other suggestions.

Relevant issue: https://github.com/neulab/explainaboard_web/issues/298


neubig (commented):
Nice!

I have a few comments. I don't think we need to resolve all of them for this PR (although I don't think it would be too hard either):

1. Chinese is a macro-language that actually covers many different language codes in ISO-639-3. You may want to consider creating a `explainaboard/utils/language_utils.py` file with the following function that detects all languages in the [Chinese macro-language family](https://iso639-3.sil.org/code/zho). I also added Japanese for good measure, and we could generalize this later:

```
# All language codes in the ISO 639-1-3 macro language family (https://iso639-3.sil.org/code/zho)
CHINESE_MACRO_FAMILY = {
  'zh',
  'zho',
  'cdo',
  'cjy',
  'cmn',
  'cpx',
  'czh',
  'czo',
  'gan',
  'hak',
  'hsn',
  'lzh',
  'mnp',
  'nan',
  'wuu',
  'yue',
  'cnp',
  'csp',
}
def is_chinese_lang_code(lang_code: str) -> bool:
  return lang_code in CHINESE_MACRO_FAMILY

def is_japanese_lang_code(lang_code: str) -> bool:
  return lang_code in {'ja', 'jpn'}
```

You can then use `is_chinese_lang_code()` to detect the Chinese language code for the tokenizer.

2. If you're going to add Chinese we might as well add Japanese and Python at the same time.

3. However, I think we need to be careful for some tasks, like sequence labeling, where we really shouldn't be changing the tokenization and `SingleSpaceTokenizer` is the right one. Right now in the implementation we have "white-list" tasks (e.g. conditional generation tasks) where we use the tokenizer, but maybe we want "black-list" tasks (e.g. sequence labeling, parsing, etc.) where we don't use the tokenizer.

odashi:
@pfliu-nlp @neubig It is better to avoid any task-specific selection form `get_default_tokenizer`: it should return task-agnostic, language-specific tokenizer to avoid unnecessary responsibility to the tasks. I think `Processor` is the one who should know which tokenizer is applicable.

```python
class FooProcessor(Processor):
    def get_overall_stats(self):
        ...
        tokenizer = self.get_tokenizer(lang)

    def get_tokenizer(self, lang):
        if lang == xx:
            return XXToeknizer()
        else:
            return tokenizer.get_default_tokenizer()
```

pfliu-nlp:
hi, @neubig great.

> regarding "is_chinese_lang_code"

do we need some code like 
```python
if is_chinese_lang_code(lang):
        return SacreBleuTokenizer(variety="zh")
```
It seems a little strange here.


@odashi 
>"It is better to avoid any task-specific selection form"

This is a good point.

neubig:
> do we need some code like

Yes, that's what I was thinking.

> "It is better to avoid any task-specific selection form"

Yes, I agree with this too, let's do it that way!

pfliu-nlp:
@neubig 
Do you assume that all `languages` in `CHINESE_MACRO_FAMILY` could be tokenized by `SacreBleuTokenizer(variety="zh")`.
It seems that SacreBleu toolkit doesn't support this.

neubig:
I guess it's a question of whether the SacreBLEU Chinese tokenizer or the SacreBLEU general purpose tokenizer is better for segmenting Min Dong Chinese, or Wu Chinese. Probably the SacreBLEU Chinese tokenizer, but I'm not a fluent Chinese speaker so I may be wrong :)

explainaboard/processors/conditional_generation.py
```diff
@@ -49,6 +53,17 @@ def task_type(cls) -> TaskType:
         """See Processor.task_type."""
         return TaskType.conditional_generation
 
+    def get_tokenizer(self, lang) -> Tokenizer:
```
odashi, line 56:
```suggestion
    def get_tokenizer(self, lang: str | None) -> Tokenizer:
```

explainaboard/processors/nlg_meta_evaluation.py
```diff
@@ -29,6 +34,17 @@ def task_type(cls) -> TaskType:
         """See Processor.task_type."""
         return TaskType.nlg_meta_evaluation
 
+    def get_tokenizer(self, lang) -> Tokenizer:
```
odashi, line 37:
```suggestion
    def get_tokenizer(self, lang: str | None) -> Tokenizer:
```

explainaboard/processors/processor.py
```diff
@@ -47,6 +47,10 @@ def task_type(cls) -> TaskType:
         """Returns the task type of this processor."""
         ...
 
+    def get_tokenizer(self, lang) -> Tokenizer:
```
odashi, line 50:
```suggestion
    def get_tokenizer(self, lang: str | None) -> Tokenizer:
```

explainaboard/processors/processor.py
```diff
@@ -47,6 +47,10 @@ def task_type(cls) -> TaskType:
...
+        """Returns a tokenizer."""
```
odashi, line 51: Add Args and Returns sections to the docstring.

explainaboard/processors/sequence_labeling.py
```diff
@@ -41,6 +42,10 @@ def __init__(self):
         super().__init__()
         self._span_ops: SpanOps = self._default_span_ops()
 
+    def get_tokenizer(self, lang) -> Tokenizer:
```
odashi, line 45:
```suggestion
    def get_tokenizer(self, lang: str | None) -> Tokenizer:
```

neubig (approved):
I approve merging the PR after @odashi 's small suggestions are reflected!

pfliu-nlp:
Thank you all for the comments!

PR #415, opened 2022-08-28
Add docstring linter

odashi:
Fixes #404

Work-in-progress.

@neubig @pfliu-nlp There are too many errors. Could you fix several files directly in this branch?

PR #488, opened 2022-09-13
Remove external_stats from MetricConfig

tetsuok:
This PR removes `external_stats` from `MetricConfig` class as the property was accidentally revived in #415.


odashi (approved):
This was a weird drawback caused by large changes... thanks!

Issue #485, opened 2022-09-13
Publish package to PyPI on push tags


tetsuok:
Currently, a GitHub action to push ExplainaBoard to PyPI is triggered on push commit to `main` if some change is maded in `version.py` (regardless of version change). This approach is not reliable because it is possible to add comments or more code to `version.py`. In fact, #415 added docstrng to `version.py` which accidentally triggered the CI, and CI failed because the version didn't change.

It is better to configure the CI not to get triggered accidentally. For example, it is possible to trigger the CI on pushing a tag (e.g., v0.12) to this repository rather than on pushing a commit to `main`. (NOTE: Whether to create release branches or not is a separate topic). Process to bump the version in `version.py`, and push tags need to be automated to avoid .

neubig:
This sounds good to me, as long as we can realize the following:

> Process to bump the version in version.py, and push tags need to be automated to avoid .

One reason why this is automated now (instead of using tags) is that in my previous experience I used tags for releases but it was error prone to need to update the code *and* the tag manually, and one was often forgotten.

odashi:
Some techniques used by other libraries is autogenerating version information from the tag: refers the tag string, and if it is formed as a version number, copy it to _version.py and publish the library with it.
There is a tool to refer information in the version control system during setup, which is used in other projects (e.g., Black)

https://pypi.org/project/setuptools-scm/

PR #492, opened 2022-09-14
Bump version to 0.11.2

OscarWang114:
Bumping to a new version so the web interface could start using the latest functionalities.

Requires #483 to be merged first.

### DIFF
--- a/version.py
+++ b/version.py
@@ -1,3 +1,3 @@
 """Version information of ExplainaBoard."""
 
-__version__ = "0.11.1"
+__version__ = "0.11.2"


odashi (approved):
Let's go ahead.

FYI: there is an issue discussing about versioning. We may apply some other practices from the next version. #485

pfliu-nlp:
@OscarWang114  
Feel free to go ahead since https://github.com/neulab/ExplainaBoard/pull/483 has been merged.
Also, one very important thing is that there are usually some potential incompatible issues in the ExplainaBoard web due to the upgrade of Explainaboard SDK. So it would be nice if you can perform some lightweight usability tests on some tasks after you upgrade the SDK version number in explainaboard_web repo.

Here is one example of the usability test: https://github.com/neulab/explainaboard_web/issues/311
You don't need to test all, maybe select some typical tasks first.

neubig:
@pfliu-nlp : FYI, we have a checklist here https://github.com/neulab/explainaboard_web/wiki/Pre-release-Checklist

You can add things to it if you'd like. It might be nice to also automate some of these things if possible.

pfliu-nlp:
Aha interesting and thanks for the pointer, I will refine it later.

OscarWang114:
Just finished class, thanks for merging the PR! 

The checklist is definitely helpful. I'll use it to check the next release and see if there's anything I'd like to add as well.

PR #493, opened 2022-09-17
including a list of samples in combo analysis result

PaulCCCCCCH:
For `ComboCountAnalysis`, we include a list of samples for each bucket. `BucketAnalysis` already has this, and we can view a list of samples when click on a bucket at the frontend. By accepting this change, we can allow the same thing for `ComboCountAnalysis`.

This PR also includes some other minor changes (doc correction and test case updates).


docs/cli_interface.md
```diff
@@ -55,12 +55,12 @@ Sentiment Treebank, a set of sentiment tags over English reviews.
 
 The below example loads the `sst2` dataset from DataLab:
 ```shell
-explainaboard --task text-classification --dataset sst2 --system-outputs ./data/system_outputs/sst2/sst2-lstm-output.txt
+explainaboard --task text-classification --dataset sst2 --system_outputs ./data/system_outputs/sst2/sst2-lstm-output.txt
```
odashi, line 58: It looks this change is wrong. We follow the GNU style for every CLI options.

https://github.com/neulab/ExplainaBoard/blob/ca23011603db2b6cfda468ba201af4dec74aaf97/explainaboard/explainaboard_main.py#L139-L148

explainaboard/analysis/analyses.py
```diff
@@ -113,6 +113,16 @@ def from_dict(dikt: dict):
                 features=tuple(dikt['features']),
             )
 
+    @staticmethod
+    def _subsample_analysis_cases(
+        sample_limit: int, analysis_cases: list[int]
+    ) -> list[int]:
```
odashi, line 119: Add docstring of this function so that other people can understand what this is.

explainaboard/analysis/analyses.py
```diff
@@ -113,6 +113,16 @@ def from_dict(dikt: dict):
...
+        if len(analysis_cases) > sample_limit:
+            sample_ids = np.random.choice(analysis_cases, sample_limit, replace=False)
+            return [int(x) for x in sample_ids]
```
odashi, line 122: Use `ndarray.tolist()` for simplicity and efficiency:
```suggestion
            return sample_ids.tolist()
```

explainaboard/analysis/analyses.py
```diff
@@ -318,7 +321,7 @@ class ComboCountAnalysisResult(AnalysisResult):
     """
 
     features: tuple[str, ...]
-    combo_counts: list[tuple[tuple[str, ...], int]]
+    combo_counts: list[tuple[tuple[str, ...], int, list[int]]]
```
odashi, line 324: 1. This variable looks no longer trivial. I think introducing a frozen dataclass is better.
2. The second member should be removed because it is fully transitive from `len(third_member)`.
```suggestion
    combo_counts: list[ComboOccurence]
```
where:
```python
@dataclass(frozen=True)
class ComboOccurence:
    """A struct representing occurences of the string tuples.

    Args:
        features: The feature values that the occurence is counted.
        sample_ids: List of sample IDs that contain the feature.
    """

    features: tuple[str, ...]
    sample_ids: list[int]
```

explainaboard/analysis/analyses.py
```diff
@@ -372,6 +375,8 @@ class ComboCountAnalysis(Analysis):
 
     features: tuple[str, ...]
     cls_name: Optional[str] = None
+    method: str = "discrete"
+    sample_limit: int = 50
```
odashi, line 379: Add descriptions of these variables in the docstring of this class.

explainaboard/analysis/analyses.py
```diff
@@ -391,11 +396,18 @@ def perform(
             if x not in cases[0].features:
                 raise RuntimeError(f"combo analysis: feature {x} not found.")
 
-        combo_map: dict[tuple[str, ...], int] = {}
+        combo_map: dict[tuple[str, ...], list] = {}
```
odashi, line 399: 1. As I commented, we don't need counts. This dict can hold only the list of IDs.
2. Defaultdict may help.

```suggestion
        combo_map: DefaultDict[tuple[str, ...], list[int]] = defaultdict(list)
```

explainaboard/analysis/analyses.py
```diff
@@ -391,11 +396,18 @@ def perform(
...
-        combo_list = list(combo_map.items())
+            if feat_vals not in combo_map:
+                combo_map[feat_vals] = [0, []]
+            combo_map[feat_vals][0] += 1
+            combo_map[feat_vals][1].append(case.sample_id)
+
+        combo_list = [
+            (k, v[0], self._subsample_analysis_cases(self.sample_limit, v[1]))
+            for k, v in combo_map.items()
+        ]
```
odashi, line 410: With the suggestions above:
```suggestion
            combo_map[feat_vals].append(case.sample_id)
        
        combo_list = [
            ComboOccurence(k, self._subsample_analysis_cases(self.sample_limit, v)
            for k, v in combo_map.items()
        ]
```

explainaboard/analysis/analyses_test.py
```diff
@@ -161,9 +161,9 @@ def test_generate_report(self) -> None:
             level="bar",
             features=("feat1", "feat2"),
             combo_counts=[
-                (("aaa", "bbb"), 123),
-                (("iii", "jjj"), 456),
-                (("xxx", "yyy"), 789),
+                (("aaa", "bbb"), 123, list(range(123))),
+                (("iii", "jjj"), 456, list(range(123, 123 + 456))),
+                (("xxx", "yyy"), 789, list(range(123 + 456, 123 + 456 + 789))),
```
odashi, line 166: It looks the test case generates very long lists that is not necessary to check the behavior.
```suggestion
                (("aaa", "bbb"), 3, list(range(3))),
                (("iii", "jjj"), 6, list(range(6))),
                (("xxx", "yyy"), 9, list(range(9))),
```

odashi (commented):
Thanks! several comments left here:

explainaboard/analysis/analyses.py
odashi: I also guessed `dict[tuple[str, ...], list[int]]` is much better.

docs/cli_interface.md
PaulCCCCCCH: Seems weird because when I run `explainaboard --task text-classification --dataset sst2 --system_outputs ./data/system_outputs/sst2/sst2-lstm-output.txt` (with underscore) it works. But when I run `explainaboard --task text-classification --dataset sst2 --system-outputs ./data/system_outputs/sst2/sst2-lstm-output.txt`, I get

```
usage: explainaboard [-h] [--task TASK] --system_outputs SYSTEM_OUTPUTS [SYSTEM_OUTPUTS ...] [--reports REPORTS [REPORTS ...]] [--systems SYSTEMS [SYSTEMS ...]] [--datasets DATASETS [DATASETS ...]] [--languages LANGUAGES [LANGUAGES ...]]
                     [--systems_aggregation SYSTEMS_AGGREGATION] [--datasets_aggregation DATASETS_AGGREGATION] [--languages_aggregation LANGUAGES_AGGREGATION] [--dataset DATASET] [--sub_dataset SUB_DATASET] [--split SPLIT] [--language TARGET_LANGUAGE]
                     [--source_language SOURCE_LANGUAGE] [--reload_stat RELOAD_STAT] [--metrics [METRICS [METRICS ...]]] [--output_file_type OUTPUT_FILE_TYPE] [--conf_value CONF_VALUE] [--output_dir OUTPUT_DIR] [--report_json REPORT_JSON] [--system_details SYSTEM_DETAILS]
                     [--custom_dataset_paths [CUSTOM_DATASET_PATHS [CUSTOM_DATASET_PATHS ...]]] [--custom_dataset_file_type CUSTOM_DATASET_FILE_TYPE]
explainaboard: error: the following arguments are required: --system_outputs
```
It somehow says it requires `--system_outputs`.

This seems a arg parser problem rather than doc problem. Will revert the change on the doc and we can fix the problem.

odashi: It looks the version of the installed binary on your machine is old for some reason. Could you try to re-install it?

PaulCCCCCCH:
> @PaulCCCCCCH is this ready to review again?

Yes. I have resolved all the issues @odashi pointed out. The unresolved conversation is about the env on my machine so it's fine. 

Thanks!

explainaboard/analysis/analyses.py
```diff
@@ -391,11 +431,16 @@ def perform(
...
-        combo_map: dict[tuple[str, ...], int] = {}
+        combo_map: defaultdict[tuple[str, ...], list[int]] = defaultdict(list)
         for case in cases:
             feat_vals = tuple([case.features[x] for x in self.features])
-            combo_map[feat_vals] = combo_map.get(feat_vals, 0) + 1
-        combo_list = list(combo_map.items())
+            combo_map[feat_vals].append(case.sample_id)
+
+        combo_list = [
+            ComboOccurence(k, self._subsample_analysis_cases(self.sample_limit, v))
```
neubig, line 440: Because `ComboOccurence` doesn't store the number of elements in the original list, limiting the number of elements in the list to `self.sample_limit` will change the result of the confusion matrix, right? I think `ComboOccurrence` probably needs another member `total_count`, which is set to `len(v)` before subsampling.

explainaboard/analysis/analyses_test.py
```diff
@@ -161,18 +162,18 @@ def test_generate_report(self) -> None:
...
             ],
         )
         report = textwrap.dedent(
             """\
             feature combos for feat1, feat2
             feat1\tfeat2\t#
-            aaa\tbbb\t123
-            iii\tjjj\t456
-            xxx\tyyy\t789
+            aaa\tbbb\t[0, 1, 2]
```
neubig, line 174: This should also be set to `ComboOccurrence.total_count`, not `ComboOccurence.sample_ids`.

explainaboard/analysis/analyses.py
```diff
@@ -318,7 +353,7 @@ class ComboCountAnalysisResult(AnalysisResult):
     """
 
     features: tuple[str, ...]
-    combo_counts: list[tuple[tuple[str, ...], int]]
+    combo_counts: list[ComboOccurence]
```
neubig, line 356:
```suggestion
    combo_occurences: list[ComboOccurence]
```

neubig (commented):
Thanks! I think the change might have introduced a bug in the confusion matrix calculation, so please check!

explainaboard/analysis/analyses.py
PaulCCCCCCH: Oh you are right! I didn't realize this. I somehow assumed `len(sample_ids)` and `total_count` would always be the same thing. Will fix this right away. Thanks!

PaulCCCCCCH:
I think we are ready to go if this looks good to everyone. Any other thoughts / comments?

explainaboard/analysis/analyses.py
```diff
@@ -307,6 +317,37 @@ def perform(
         )
 
 
+@dataclass(frozen=True)
+class ComboOccurence:
+    """A struct representing occurences of the string tuples.
+
+    Args:
+        features: The feature values that the occurence is counted.
+        sample_ids: List of sample IDs that contain the feature.
```
odashi, line 326:
```suggestion
        sample_count: Number of occurrences of the feature.
        sample_ids: List of sample IDs that contain the feature.
            This list may contain subsampled IDs to suppress memory
            efficiency, so `len(sample_ids) <= sample_count`.
```

explainaboard/analysis/analyses.py
```diff
@@ -307,6 +317,37 @@ def perform(
...
+    def from_dict(dikt: dict) -> ComboOccurence:
+        """Deserialization method."""
+        return ComboOccurence(
+            features=dikt['features'],
+            sample_count=dikt['sample_count'],
+            sample_ids=dikt['sample_ids'],
+        )
+
+    def __lt__(self, other: ComboOccurence):
+        """Implement__lt__ to allow natural sorting."""
```
odashi, line 343:
```suggestion
    def __lt__(self, other: ComboOccurence) -> bool:
        """Implement __lt__ to allow natural sorting."""
```

explainaboard/analysis/analyses.py
```diff
@@ -307,6 +317,37 @@ def perform(
...
+        return (self.features, self.sample_count, self.sample_ids) < (
+            other.features,
+            self.sample_count,
+            other.sample_ids,
+        )
```
odashi, line 348: Randomized variables is not suitable for comparison, and it may cause unintended behavior.

```suggestion
        return (self.features, self.sample_count) < (
            other.features,
            other.sample_count
        )
```

odashi (commented):
Several fixes on new changes.

explainaboard/analysis/analyses.py
PaulCCCCCCH: Resolved. Thanks a lot! 

Just out of curiosity, why would that be a bug?

odashi: > why would that be a bug?

Because the RHS contained `self.sample_count`

explainaboard/analysis/analyses.py
```diff
@@ -339,13 +342,9 @@ def from_dict(dikt: dict) -> ComboOccurence:
...
-    def __lt__(self, other: ComboOccurence):
-        """Implement__lt__ to allow natural sorting."""
-        return (self.features, self.sample_count, self.sample_ids) < (
-            other.features,
-            self.sample_count,
-            other.sample_ids,
-        )
+    def __lt__(self, other: ComboOccurence) -> bool:
+        """Implement __lt__ to allow natural sorting."""
+        return (self.features, self.sample_count) < (other.features, self.sample_count)
```
odashi, line 347:
```suggestion
        return (self.features, self.sample_count) < (other.features, other.sample_count)
```

explainaboard/analysis/analyses.py
PaulCCCCCCH: Fixed. Thanks!

neubig (approved):
Looks good to me now!

odashi (approved):
Thanks!

PR #499, opened 2022-09-19
Strict shape checking of `aggregate_stats` and `calc_metric_from_aggregate`

### LINKED ISSUE #497: aggregate_stats and calc_metric_from_aggregate don't work in some cases.
`Metric.aggregate_stats` and `Metric.calc_metric_from_aggregate` does not guarantee that each implementation returns the ndarray with correct shape. These functions **must** return following arrays:

- `aggregate_stats` ... `[num_stats]` or `[num_batches, num_stats]`
- `calc_metric_from_aggregate` ... `[]` or `[num_batches]`

But several implementations does return other shapes, even the default implementations in `Metric`.

This causes several wrong consequences. A serious one is bootstrapped CI never returns correct data because the inner `sort()` doesn't work along the correct axis (I observed this when adding a unit test for `calc_confidence_interval`).

odashi:
This change introduces strict shape checking of return values of `Metric.aggregate_stats` and `Metric.calc_metric_from_aggregate`. These functions are now defined as a wrapper of inner functions (that may be overridden), and checks if the return values have expected shape.

May fix #497.


odashi:
@neubig This change generates a bunch of errors due to unclear ndarray operations in inherited classes. I couldn't fully realize the correct fix of each implementation. Could you take a look at and fix them if possible?

neubig:
Thanks @odashi ! I fixed most.

@pfliu-nlp : it seems that the remaining ones are APE and NLG meta-evaluation, where you're more familiar than I am. Would you be able to take a look? My commit here might be a good reference: https://github.com/neulab/ExplainaBoard/pull/499/commits/de3dac48b5cffcadf99d1980af4cddad44bc22d0

pfliu-nlp:
Sure. I'm working on this.

explainaboard/metrics/metric.py
```diff
@@ -269,44 +269,107 @@ def calc_stats_from_data(
...
+                - Non-batched data: [num_stats]
+                - Batched data: [batch_size, num_stats]
         """
+        result = self._aggregate_stats(stats)
+
+        if stats.is_batched():
+            stats_shape = stats.get_batch_data().shape
+            result_shape: tuple[int, ...] = (stats_shape[0], stats_shape[2])
+        else:
+            result_shape = (stats.num_statistics(),)
```
pfliu-nlp, line 303: @odashi Could you tell me why this is the case? i.e., Must one of the dimensions of aggregated result be the same as `num_statistics`?

odashi: Yes (see docstring). This function expects to process only aggregating the same statistics ([[A,B,C], [A',B',C']] -> [A'', B'', C''])

pfliu-nlp: I see. 

So there are two types of `aggregation`s,
* (1) aggregation over different samples
* (2) aggregation over different statistics 

Based on my understanding, the current aggregation function says: (1) must happen before (2).
I'm not sure if this is too strong. 
For example, there are some scenarios, where (2) should happen first (different statistics should be composed together to calculate a metric value)

What do you think of this situation?

neubig: Hi @pfliu-nlp : just to clarify, are you referring to the following?
* (1) `aggregate_stats()`
* (2) `calc_metric_from_aggregate()`

If that's the case, `calc_metric_from_aggregate()` can also be applied to batched statistics, so it could theoretically be called before `(1)` as well to calculate metrics on multiple examples as well.

odashi: Some confusion comes from the fact that the interface takes and returns ndarrays directly.
More concretely, we need to make sure that:

1. `aggregate_stats` takes the concept *Stats* and returns the concept *Aggregate*
2. `calc_metric_from_aggregate` takes the concept *Aggregate* and retruns the concept *Result*

*Stats*, *Aggregate* and *Result* are represented as `ndarray`s (though *Stats* are obtained from `MetricStats`), but they don't share any underlying semantics.
According to this understanding, there is no chance that 2. precedes 1 as @neubig said too.

odashi: More and more concretely:

- $\mathrm{aggregate\\_stats}: List[Stats] \rightarrow Stats \ \ OR \ \ Batch[List[Stats]] \rightarrow Batch[Stats]$
- $\mathrm{calc\\_metric\\_from\\_aggregate}: Stats \rightarrow Result \ \ OR \ \ Batch[Stats] \rightarrow Batch[Result]$

neubig: Yes, I like this formulation.

odashi:
According to the internal discussion, we will introduce `use_customized_aggregate()` flag function that informs the assertion to skip to check the size of the last dimension.

neubig:
I'll take a quick look at this.

neubig:
@odashi and @pfliu-nlp : tests seem to be passing. you could take a look at my commits and see if you have comments.

PR #502, opened 2022-09-21
Remove prop_samples from calc_confidence_interval

tetsuok:
The sampling part of the procedure of *Bootstrapping* can be summarized at high level if I understand it correctly:
1. Resample each sample in the given data.
2. Repeat Step 1 several times

where the number of random samples generated in Step 1 should be same as the number of samples in the data. 
(The similar description is found in [scipy.stats.bootstrap](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.bootstrap.html#scipy-stats-bootstrap))

However, it seems `calc_confidence_interval` implements differently, which raises a concern for the correctness of the algorithm. I could be wrong, but it seems better to prefer the correctness over the efficiency of resampling given that computed values with this library could be used in someone's research. With this in mind, this PR removes `prop_samples` from `calc_confidence_interval`.

### DIFF
--- a/explainaboard/metrics/metric.py
+++ b/explainaboard/metrics/metric.py
@@ -364,11 +364,11 @@ class Metric:
             )
         # Do bootstrapping otherwise
         else:
-            n_elems = len(stats)
+            sample_size = len(stats)
             all_indices = np.array(range(len(stats)))
             rng = np.random.default_rng()
             all_indices = rng.choice(
-                all_indices, size=(n_samples, n_elems), replace=True
+                all_indices, size=(n_samples, sample_size), replace=True
             )
             filt_stats = stats.filter(all_indices)
             agg_stats = self.aggregate_stats(filt_stats)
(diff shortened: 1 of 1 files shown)


odashi:
FYI: We can find some explanations, e.g., https://www.stata.com/support/faqs/statistics/bootstrapped-samples-guidelines/

Roughly speaking, resampling from $n$ data with other sample size $n'$ changes the tendency of obtained statistics from the original sample. To avoid this effect, we need to always use the same sample size $n$ for every resampling round, or the obtained statistics lacks the correct meaning.

explainaboard/metrics/metric.py
```diff
@@ -366,7 +364,7 @@ def calc_confidence_interval(
             )
         # Do bootstrapping otherwise
         else:
-            n_elems = max(int(prop_samples * len(stats)), 1)
+            n_elems = len(stats)
```
odashi, line 367: I think the name of this variable should be `sample_size`.
And it can be used too in the next line.

tetsuok: Done. Rename `n_samples` to `num_iterations` to avoid confusion with `sample_size`.

odashi (approved):

Issue #482, opened 2022-09-12
`value` should be set to `None` under empty buckets


OscarWang114:
As mentioned in #476, `value` should be set to `None` under empty buckets. This requires a fix on both SDK and the web interface.

odashi:
Maybe this issue will be resolved if we applied #477 and skip generating results for empty buckets.

PR #503, opened 2022-09-23
Implement MetricValues

### LINKED ISSUE #477: Decompose MetricResult
`MetricResult` has several information that are basically independent each other:

- A "value" with unspecified meaning.
- A confidence interval of "value"
- An agreement "value" (comes from `AuxiliaryMetricResult` which is used only to store Fleiss' Kappa).

Especially, there is an ambiguity between the "value" and auxiliary scores: we don't have any consensus to decide which number is "the value" of the metric.

To alleviate this confusion, I thought that `MetricResult` can be constructed as a set of values instead of having every field by itself, represented as follows:

```python
class MetricResult:
    config: MetricConfig
    values: dict[str, MetricValue]
```

where each `MetricValue` represents an individual statistical concept of the metric:

```python
class MetricValue:
    pass

class Score(MetricValue):
    value: float

class ConfidenceInterval(MetricValue):
    low: float
    high: float
    alpha: float
```

Using these definitions, we can represent the results as follows (for example):

```python
# Metric with agreement
MetricResult(
    config=config,
    values={
        "score": Score(value),
        "score_ci": ConfidenceInterval(low, high, alpha),
        "agreement": Score(agreement),
    },
)
```

This design has several benefits:

- We no longer need to determine which is the "value" of the metric. Each score has a distinctive name.
- We no longer need optional fields. Absence of the score is represented as just "no entry".
- Extensibility: it supports any number of results, and we can add new concepts by implementing a new `MetricValue`.
- Type information of `values` (`Score`, `CI` in this case) explicitly represents the concept of value.

RFC: @neubig @pfliu-nlp

odashi:
Resolves #477.

This change introduces `MetricValue` classes to represent some categories of `MetricResult` values. Currently there are `Score` (holds a scalar value) and `ConfidenceInterval` (holds low/high/alpha values).
`MetricResult` is restructured to hold `dict[str, MetricValue]`.

This change also moves some unit tests related to this change form `integration_tests` to `explainaboard`.


odashi:
@OscarWang114 I request your review too as the representative of #482

neubig (approved):

PR #505, opened 2022-09-25
Refactor preprocessor.py

odashi:
This change refactors `explainaboard/utils/preprocessor.py`. Specifically, this removes `Preprocrssor` class and all subclasses are reformed to behave as `Callable[[str], str]`.

### DIFF
--- a/explainaboard/loaders/file_loader.py
+++ b/explainaboard/loaders/file_loader.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable
 import copy
 import csv
 import dataclasses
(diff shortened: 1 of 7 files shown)


pfliu-nlp:
Thanks for this PR! A high-level question:
It seems that this PR basically reduces the inheritance behavior of `Preprocessor`, for example
previously we have 
* `Preprocessor -> MapPreprocessor` 
* `Preprocessor -> ExtractiveQAPreprocessor` 
After refactoring
* `MapPreprocessor` 
* `ExtractiveQAPreprocessor` 

Also, we need to update the old typing `Preprocessor` -> `Callable[[str], str]`.

Just wondering what the general principle is behind such refactoring? For example, complicated inheritance makes it hard to understand.

neubig (approved):

odashi:
@pfliu-nlp I assumed that the interface looks premature since the `Preprocessor` provides nothing more than `Callable[[str], str]`. If we need some other functionality on it, we can implement the interface again.

PR #508, opened 2022-09-27
Bump mypy version to 0.981

tetsuok:
This PR bumps mypy version to 0.981. This is to remove "ignore" comments to silence mypy. 

* The version includes a fix for python/mypy#5374. This PR removes the comments to silence mypy due to the issue.
* The version has a fix for python/mypy#731. The fix is enabled when running mypy with `--enable-recursive-aliases` ([link to blog post](http://mypy-lang.blogspot.com/2022/09/mypy-0981-released.html)), but this PR doesn't enable the fix, deferring to a separate PR. This is because using the option causes different 91 errors. It seems better to tackle enabling the option in a separate PR.
* The version doesn't have a fix for python/mypy#4717, unfortunately. mypy still complains when the comments were removed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/python/black.git
-    rev: 22.3.0 
+    rev: 22.3.0
     hooks:
       - id: black
         files: '\.py$'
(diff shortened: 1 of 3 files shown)


odashi:
@tetsuok Maybe this update resolves some issues tagged by 4717 as well as 5374.

odashi (commented):
dummy

tetsuok:
@odashi Issues on python/mypy#4717 still persist with this update.

odashi (approved):

Issue #491, opened 2022-09-14
Remove MetricConfig and directly store configs in Metric


odashi:
`Metric` is instantiated by appropriate `MetricConfig`. This strategy is used only in `Metric` and any other classes are instantiated by directly calling `__init__` with specific arguments. This special architecture around `Metric` actually causes some confusion (e.g., #455).
This separation also requires explicit circular dependency: `Metric` holds `MetricConfig`, while `MetricConfig` knows corresponding `Metric` (in `to_metric`). This is somewhat annoying because it always lets us to refactor both concepts at the same time.

As long as I checked the all subclasses of `Metric`, there is no heavy implementations: they simply holds configs without any special initialization processes. It looks we have no any reason to separate `MetricConfig` from `Metric` at interface level. Integrating them into the `Metric` could simplify the architecture.

RFC: @neubig @pfliu-nlp @tetsuok

odashi:
`MetricConfig` is stored in `MetricResult` too, but it seems this information is not necessarily required. There is also `Performance` that has almost the same information with `MetricResult`, and the only difference is that it stores metric's name instead.
I think using `Performance` rather than `MetricResult` is sufficient. We can pass the original Metric with the corresponding `Performance` if the both should be referred.

tetsuok:
@odashi I agree with the proposal and the potential changes. Removing the circular dependency seems a good idea to make things simpler.

pfliu-nlp:
@odashi 
> "Remove MetricConfig, store every information necessary to perform metrics to Metric."

In general, I think it looks good. I think the reason why we have two separate classes originally (`Metric` and `MetricConfig`) is to handle the following case (ref: https://github.com/neulab/ExplainaBoard/blob/cc2d315695cec30e21d369362d1c76ca2ce9d40c/explainaboard/processors/kg_link_tail_prediction.py#L134)

```python
            HitsConfig(name='Hits1', hits_k=1),
            HitsConfig(name='Hits2', hits_k=2),
            HitsConfig(name='Hits3', hits_k=3),
            HitsConfig(name='Hits5', hits_k=5),
            HitsConfig(name='Hits10', hits_k=10),
```
When we merge `MetricConfig` into `Metric`, I think we will have two names for each `Metric` class: `metric_name` and `metric_class_name`.

(Again, I think this refactoring will potentially affect explainaboard web and I will follow PRs on this discussion.)

odashi:
[...]

PR #511, opened 2022-09-27
Introduce dict[str, Performance]

odashi:
Related to #491

This PR introduces following changes:

- Remove `metric_name` member from `Performance`.
- Change all `list[Performance]` to `dict[str, Performance]`, whose keys are metric names.
- Refactoring related to these changes.

Originally, `metric_name` is used as a primary key of the list, and often the list is converted to dict in the library. This change attempts to removes this burden.

### DIFF
--- a/explainaboard/analysis/analyses.py
+++ b/explainaboard/analysis/analyses.py
@@ -163,20 +163,15 @@ class BucketAnalysisResult(AnalysisResult):
 
     def __post_init__(self):
         """Set the class name and validate."""
-        metric_names = [x.metric_name for x in self.bucket_performances[0].performances]
-        num_metrics = len(metric_names)
+        metric_names = self.bucket_performances[0].performances.keys()
+
         for bucket_perf in self.bucket_performances:
-            if len(bucket_perf.performances) != num_metrics:
+            if bucket_perf.performances.keys() != metric_names:
                 raise ValueError(
-                    "Inconsistent number of metrics. "
-                    f"Required: {num_metrics}, got: {len(bucket_perf.performances)}"
+                    "Inconsistent metrics. "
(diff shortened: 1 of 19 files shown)


explainaboard/analysis/analyses.py
```diff
@@ -269,12 +269,8 @@ def perform(
...
 
             n_samples = len(bucket_collection.samples)
-            bucket_performance = BucketPerformance(
-                n_samples=n_samples,
-                bucket_samples=subsampled_ids,
-                bucket_interval=bucket_collection.interval,
-                bucket_name=bucket_collection.name,
-            )
+
+            performances: dict[str, Performance] = {}
```
pfliu-nlp, line 268: Do we need to require that the `performances` need to maintain insertion order?

pfliu-nlp (approved):

pfliu-nlp:
One high-level comment that could be considered as future PRs.
* refactoring `metric_configs`, `overall` and  `analyses` to dict[str, ...] as well

explainaboard/analysis/analyses.py
neubig: It would be nice to maintain this property. It seems that as of Python 3.7, dicts are guaranteed to maintain insertion order: https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6

neubig (approved):

explainaboard/analysis/analyses.py
odashi: I don't think assuming any order is semantically good idea if we are using hashmap (dict) even if the language supports it. If the order of names is meaningful for someone, arranging the order should be the responsibility of callers.

PR #512, opened 2022-09-28
Remove additional config argument from Metric

odashi:
This change removes `config` arguments from methods in `Metric`.

Overall, this argument is rarely utilized while it is not referred appropriately in many subclasses. Since we can instantiate another `Metric` object to use different configs, I thought that we can remove this argument without any drawbacks.

This pull request also contains some typing errors in `Metric` classes.

### DIFF
--- a/explainaboard/metrics/accuracy.py
+++ b/explainaboard/metrics/accuracy.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Optional
+from typing import Any
 
 import numpy as np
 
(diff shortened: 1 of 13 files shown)


pfliu-nlp (approved):
This is pretty straightforward and looks good, I approved.

neubig (approved):
I approve.

Just to clarify, the reason why this was implemented this way in the first place was because we were assuming that there may be some evaluation metrics (e.g. embedding-based metrics) where the initialization cost is high, but metrics with different configurations (e.g. different languages to be evaluated) could be run with the same initialization. However, this has become unimportant given how we use EB now (basically we re-start the program every time we evaluate a new system).

odashi:
@neubig I think we should implement some inner singleton to hold sharable objects in the same metric instead of overriding the config.

neubig:
Sounds good, I'd suggest we revisit that if when it becomes necessary.

PR #513, opened 2022-09-29
Remove MetricResult.config.

odashi:
Related to #491

This change removes `MetricResult.config`. This member is not used at this point, and we basically don't need to store the same config in this struct because we can obtain the same information directly from `Metric`.


neubig (commented):
Quick clarification: When writing out the results of an analysis, I believe only `MetricResult` is serialized, not `Metric`. How would it be possible to get the information about `Metric` when loading in the results from a serialized file?

odashi:
I think `MetricConfig` (or `Metric`) should be serialized separately as well. I also think parallel structure is more maintainable than nested structure.

```json
"metric_configs": {
  "foo": <foo_config>,
  "bar": <bar_config>
},
"metric_results": {
  "foo": <foo_result>,
  "bar": <bar_result>
}
```

neubig:
OK, sounds good. I'll review this.

neubig (approved):

PR #514, opened 2022-09-30
Remove unnecessary Optional

odashi:
This change removes several `Optional` and `| None` annotations from `list` and `dict` types that can represent nothing by the empty collection. This change suppresses some `unwrap` invocations from the codebase.

This change also involves several typing fixes.

### DIFF
--- a/docs/example_scripts/test_kg.py
+++ b/docs/example_scripts/test_kg.py
@@ -18,8 +18,8 @@ data = loader.load()
 processor = get_processor_class(TaskType.kg_link_tail_prediction)()
 sys_info = processor.process(metadata={}, sys_output=data.samples)
 
-fine_grained_res = unwrap(sys_info.results.analyses)
-overall_res = unwrap(sys_info.results.overall)
+fine_grained_res = sys_info.results.analyses
+overall_res = sys_info.results.overall
 
 # print bucket information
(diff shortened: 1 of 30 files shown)


neubig:
Thank you! I think this is a reasonable change, but I do feel that the semantics of none and the empty list or dictionary are a little bit different in some of these cases. None generally refers to " has not been calculated yet " we're an empty list or dictionary more indicates that it has been calculated but is empty. Do you think it would be better to remove this implicit distinction? Of course we could make it explicit where necessary, but that adds a bit of complexity to the code

odashi:
> None generally refers to " has not been calculated yet " we're an empty list or dictionary more indicates that it has been calculated but is empty.

I'm not sure this is really a general policy in Python, but I don't recommend giving any other meanings than "never" to None. It is actually confusing because the value does not say anything about the underlying semantics, and we need careful documentation about it.

odashi:
I think `SysOutputInfo` (for example) has too many information that are not necessary to be co-existed. Giving/passing this structure seems to confuse overall behavior of the library.

neubig:
OK, sounds good, let's go ahead and remove the `None`s then. I'll do the review soon.

pfliu-nlp (approved):
LGTM

neubig (approved):

Issue #173, opened 2022-03-25
Unsafe en_core_web_sm downloading in setup.py


odashi:
Currently `setup.py` will execute an external command `python -m spacy download en_core_web_sm` to install a spaCy model during setup. This approach has several issues about system consystency:

- spaCy models are intendedly not registered to PyPI, and PyPI does not allow libraries depending on external requirements.
- The command is just a system command which possibly breaks the system, or won't work correctly.

Since there is no recommended way to add spaCy models to `install_requires`, we need to take either of follows:

- Download the model programatically when `spacy.load()` fails.
- Bundle the model file into this repository.
- Ask users to download appropriate models additionally.

odashi:
Some comments:
- Download the model programatically when `spacy.load()` fails.
  - This is easy option for novices, but the system is still changed unexpectedly. This option won't work in offline systems.
- Bundle the model file into this repository.
  - This is also easy yo use, but the size of the repository or distribution packages grows, and we have responsibility to maintain bundled models.
- Ask users to download appropriate models additionally.
  - This is desirable for the ease of maintenance, but users are required to type an additional command.

odashi:
Some comments? @pfliu-nlp @neubig

neubig:
I'm not 100% sure I understand the issue here actually.
So I can understand better: what is an example of a situation where the current method causes a problem for users?

neubig:
Also, as a meta-comment: currently we are planning that majority of users of the SDK in this repo will be developers or power-users who want to actually program new analyses.

We expect that most end-users will interact mostly with the web interface, and won't have to install this repo. Because of this, the emphasis on this repo being extremely easy for everyone to install is not quite as high as it may be for some other consumer software. That being said, it shouldn't be *hard* of course :)

odashi:
It is generally a strong assumption that every system command can work as we implicitly expected. for example:

1. A case we could meet easily is that the "python" command is not supported officially by several operating systems, e.g., recent versions of Ubuntu requires additional installs.
2. Some build systems (e.g., Bazel) hacks the system commands to process tool specific things. In this case the command someone invoked is sometimes not the command we expected to be run.

In addition, since the spacy model is additionally installed by the external command, it is out of dependency management at installation. This means that pip could not resolve version collisions when some other tools also requires the spacy model.

Depending on external resources from PyPI packages is not viable according to their policy:
https://pip.pypa.io/en/stable/news/#id666
so we need to avoid such installation if we use PyPI.

[...]

PR #517, opened 2022-10-01
Remove downloading spaCy's English pipelines from setup.py

### LINKED ISSUE #490: Use setup.cfg to store every setup config
Current `setup.py` follows a legacy pattern: writing settings in the code. Recent versions of setuptools supports to store most configs into `setup.cfg`, and we don't need writing setup "script" anymore.
This can avoid some complicated issues (typically circular imports) around `__version__`. `setup.cfg` can import/evaluate version file without any effects against code base.

The only issue here is the installation of `en_web_core_sm` spaCy package, which is performed at the last step of the `setup.py`. As I noted in #173, this is not a reliable way, and manually installing this package by users is much better than doing too much by us.

The underlying task in this issue is follows:

- [x] Move every setting in `setup.py` to `setup.cfg`.
- [ ] Remove ad-hoc installation of `en_web_core_sm` and add documentation about manual installation of it.

tetsuok:
This PR removes downloading spaCy's English pipelines, `en_core_web_sm` from `setup.py`. This also corrects the processor which actually depends on `en_core_web_sm`. 

Fixes #490.


neubig (approved):

odashi:
This also fixes #173

odashi (approved):

PR #519, opened 2022-10-02
Fix confidence level in scipy.stats.t.interval.

### LINKED ISSUE #510: t-test confidence level is wrong
I found that the `alpha` parameter of the `scipy.stats.t.interval` **actually takes the exact confidence level $(1 - \alpha)$**.

無題

It looks the generated interval is wrongfully tight.
https://github.com/neulab/ExplainaBoard/blob/cc2d315695cec30e21d369362d1c76ca2ce9d40c/explainaboard/metrics/metric_test.py#L175-L180

odashi:
Fixes #510.


neubig (approved):
Thanks for catching this!

PR #521, opened 2022-10-02
Implement Loader factory

odashi:
Fixes #421 

This pr implements a factory to obtain Loader classes, and removes the loader registry.

PR #434, opened 2022-08-31
Use common registry to register processors

tetsuok:
This PR changes the implementation of Processor registry from its own implementation to the common registry.
With this change, every processor class is registered by its class name, rather than `TaskType` based on the [suggestions](https://github.com/neulab/ExplainaBoard/pull/434#issuecomment-1255500133). This also adds a replacement of `get_processor`, `get_processor_class`, removing `get_processor` ([link to the summary of the discussion](https://github.com/neulab/ExplainaBoard/pull/434#discussion_r981119771)).

Fixes #422.

PR #522, opened 2022-10-02
Remove serializable from processors

odashi:
This is a follow-up change of #422.

Processors are not serialized in any data. This change removes the `Serializable` interface from all processors to avoid unnecessary confusion in the common registry.


explainaboard/processors/__init__.py
```diff
@@ -3,54 +3,6 @@
 When a new processor is implemented, remember to import it here so it gets registered.
```
neubig, line 3: This comment can be removed I think.

neubig (approved):

odashi:
@tetsuok
I noticed that there are several bugs around the registry in #521, and similar bugs may be involved in the future in the Processor's factory too.
I decided to implement unit tests around Processors as well as you first implemented in #434 regardless of its cost. Please take a look at this change to check if they are implemented appropriately.

tetsuok (approved):
LGTM

neubig (approved):

PR #523, opened 2022-10-02
fix-bug-nlg-metaevaluation

pfliu-nlp:
There are some bugs in the meta-evaluation task, and the calculated overall results are wrong. After some advice from @qinyiwei , this PR is finished:
* fix bugs
* add strict test

### DIFF
--- a/explainaboard/metrics/nlg_meta_evaluation.py
+++ b/explainaboard/metrics/nlg_meta_evaluation.py
@@ -188,10 +188,21 @@ class KtauCorrelation(CorrelationMetric):
         num = 0
         for i in range(1, len(score)):
             for j in range(0, i):
-                if abs(score[i][0] - score[j][0]) >= config.threshold:
-                    manual_better = score[i][0] > score[j][0]
-                    system_better = score[i][1] > score[j][1]
-                    if manual_better == system_better:
+                if (
+                    abs(score[i][0] - score[j][0]) < config.threshold
+                    or abs(score[i][0] - score[j][0]) == 0
+                ):
+                    continue
+                elif (
+                    score[i][0] - score[j][0] >= config.threshold
+                ):  # system i is better than system j
+                    if score[i][1] > score[j][1]:
+                        conc += 1
+                    else:
+                        disc += 1
+                    num += 1
+                else:  # system i is worse than system j
+                    if score[i][1] < score[j][1]:
                         conc += 1
                     else:
                         disc += 1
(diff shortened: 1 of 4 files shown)


explainaboard/metrics/nlg_meta_evaluation.py
```diff
@@ -188,10 +188,21 @@ def _count(self, score: list, config: Optional[MetricConfig] = None):
...
+                elif (
+                    score[i][0] - score[j][0] >= config.threshold
+                ):  # system i is better than system j
+                    if score[i][1] > score[j][1]:
+                        conc += 1
+                    else:
+                        disc += 1
+                    num += 1
+                else:  # system i is worse than system j
+                    if score[i][1] < score[j][1]:
```
odashi, line 205: It looks the same stats are used multiple times. Could you apply following variables to judge the conditions?

- `manual_diff = score[i][0] - score[j][0]`
- `system_diff = score[i][1] - score[j][1]`

explainaboard/metrics/nlg_meta_evaluation.py
```diff
@@ -234,19 +245,27 @@ class PearsonCorrelation(CorrelationMetric):
...
+        elif config.group_by == "system":
+            for _, group_vals in scores.items():
+                manual_scores = [val[0] for val in group_vals]
+                system_scores = [val[1] for val in group_vals]
+                if len(manual_scores) == 0 or len(system_scores) == 0:
+                    continue
+                manual_score.append(sum(manual_scores) / len(manual_scores))
+                system_score.append(sum(system_scores) / len(system_scores))
+        else:
+            raise NotImplementedError
```
odashi, line 267: ValueError with an appropriate message is better.

integration_tests/nlg_meta_eval_test.py
```diff
@@ -36,6 +36,6 @@ def test_da_cs_en(self):
         processor = get_processor_class(TaskType.nlg_meta_evaluation)()
 
         sys_info = processor.process(metadata, data)
-
-        self.assertGreater(len(sys_info.results.analyses), 0)
-        self.assertGreater(len(sys_info.results.overall), 0)
+        self.assertAlmostEqual(
```
odashi, line 39: I think the original assertion does some different checks, and needs to be maintained as well.

explainaboard/metrics/nlg_meta_evaluation.py
pfliu-nlp: good suggestion

explainaboard/metrics/nlg_meta_evaluation.py
```diff
@@ -188,21 +188,20 @@ def _count(self, score: list, config: Optional[MetricConfig] = None):
         num = 0
         for i in range(1, len(score)):
             for j in range(0, i):
-                if (
-                    abs(score[i][0] - score[j][0]) < config.threshold
-                    or abs(score[i][0] - score[j][0]) == 0
-                ):
+                manual_diff = score[i][0] - score[j][0]
+                system_diff = score[i][1] - score[j][1]
+                if abs(manual_diff) < config.threshold or abs(manual_diff) == 0:
```
odashi, line 193: I noticed that the following is much simpler:
```python
if manual_diff >= config.threshold:
   ...  # i is better than j
elif manual_diff <= -config.threshold:
   ...  # i is worse than j
#that's all
```

odashi (approved):
Approves it, but there is one comment

neubig (approved):

explainaboard/metrics/nlg_meta_evaluation.py
pfliu-nlp: Thanks. Fixed.

PR #526, opened 2022-10-02
Introduce meta eval metric for general NLG tasks

pfliu-nlp:
Previously, we have meta-evaluation metrics (and tasks) specifically defined for WMT, which are too dataset dependent and not generalized enough for other NLG tasks, such as summarization and data-to-text. This PR aims to introduce meta-evaluation metrics for generation tasks such as summarization and data-to-text.

PR #527, opened 2022-10-02
Add task metaevaluation for nlg

pfliu-nlp:
Blocked by: https://github.com/neulab/ExplainaBoard/pull/526

Based on evaluation metrics achieved in PR 526, this PR aims to introduce task processor.

Notably,
the shape check process in function `aggregate_stats()` is still too strong: https://github.com/neulab/ExplainaBoard/blob/b31d5d6506bdb6fb633b836ef798f25488f4052d/explainaboard/metrics/metric.py#L406 I further relax it in this PR. We can discussion more about this.


neubig:
Comment: I wonder if you'd be able to use a method for aggregation like I did here?
https://github.com/neulab/ExplainaBoard/blob/main/explainaboard/metrics/nlg_meta_evaluation.py#L115-L141

It seems that it would then be possible to avoid removing that check.

explainaboard/loaders/meta_evaluation_nlg.py
```diff
@@ -0,0 +1,60 @@
...
+from explainaboard.loaders.loader import Loader
+from explainaboard.loaders.loader_registry import register_loader
+
+
+@register_loader(TaskType.meta_evaluation_nlg)
+class MetaEvaluationNLGLoader(Loader):
+    """Loader for the natural language generation task.
+
+    usage:
+        please refer to `test_loaders.py`
```
odashi, line 22: This docstring looks redundant.

explainaboard/loaders/meta_evaluation_nlg.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @classmethod
+    def default_dataset_file_type(cls) -> FileType:
+        """See Loader.default_dataset_file_type."""
+        return FileType.datalab
+
+    @classmethod
+    def default_dataset_file_loaders(cls) -> dict[FileType, FileLoader]:
+        """See Loader.default_dataset_file_loaders."""
+        target_field_names = ['source', 'references', 'hypotheses', 'manual_scores']
```
odashi, line 33: Could you avoid using single quotes? mixing single/double quotes lacks consistency in the repository (especially, it reduces searchability of IDEs)

explainaboard/loaders/meta_evaluation_nlg.py
```diff
@@ -0,0 +1,60 @@
...
+        return {
+            FileType.datalab: DatalabFileLoader(
+                [
+                    FileLoaderField("source_column", target_field_names[0], str),
+                    FileLoaderField("references_column", target_field_names[1], list),
+                    FileLoaderField("hypotheses_column", target_field_names[2], dict),
+                    FileLoaderField("scores_column", target_field_names[3], list),
```
odashi, line 40: target_field_names is used only here: it is better to use literals instead of arrays.

explainaboard/loaders/meta_evaluation_nlg.py
```diff
@@ -0,0 +1,60 @@
...
+        return FileType.json
+
+    @classmethod
+    def default_output_file_loaders(cls) -> dict[FileType, FileLoader]:
+        """See Loader.default_output_file_loaders."""
+        field_name = "auto_scores"
+        return {
+            FileType.json: JSONFileLoader(
+                [
+                    FileLoaderField("auto_scores", field_name, list),
```
odashi, line 57: ditto: it is enough to use literals rather than a variable.

odashi (changes requested):
You need to separate changes in metric.py from this pull request. This change requires another discussion.

Also, I don't agree with the changes in metric.py at this point because of not sufficient explanation. Could you clarify more about why the change is necessary?

pfliu-nlp:
> Comment: I wonder if you'd be able to use a method for aggregation like I did here? https://github.com/neulab/ExplainaBoard/blob/main/explainaboard/metrics/nlg_meta_evaluation.py#L115-L141
> 
> It seems that it would then be possible to avoid removing that check.

I think about this but don't think it could work since once we perform
```
data.reshape((data.shape[0], data.shape[-2] * data.shape[-1]))
```

@neubig It's hard for us to recover the data to its original shape. (In the above case, `data.shape[-1]` has fixed dimension and is hard-coded as `4`. In the current case, it's dynamic.  

We need to figure out a way to fix this since it also blocks the PR: https://github.com/neulab/ExplainaBoard/pull/526.

odashi:
@pfliu-nlp For a quick fix, you can also store the size of the dimension as another stats.

pfliu-nlp:
> @pfliu-nlp For a quick fix, you can also store the size of the dimension as another stats.

Yeah, that would be another solution we can consider. But if following this, I feel like the function `_aggregate_stats` and `calc_stats_from_data` have been hacked too much. How do you think @neubig

odashi:
@pfliu-nlp Yes, it is definitely a hack, but it looks better to mitigate a restriction of the methods' presuppositions.

explainaboard/metrics/nlg_meta_evaluation.py
```diff
@@ -3,10 +3,10 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Optional, Union
+from typing import Any, Optional, Union
```
odashi, line 6: I think this file should be renamed to `meta_evaluation_nlg.py` to keep consistency with other files.

Issue #417, opened 2022-08-29
Implement CalibrationAnalysis


neubig:
Calibration is whether a system's confidence is well-correlated with whether the system got the answer right or not. It would be nice if we could do analyses related to calibration, such as calculating expected calibration error: https://arxiv.org/abs/1706.04599

I think this should probably be implemented as an additional variety of analysis, which would be simple and self-contained: https://github.com/neulab/ExplainaBoard/blob/main/explainaboard/analysis/analyses.py#L45

pfliu-nlp:
@neubig, this is also one feature I have been expecting.  The only complicated thing is we need `probability` as an additional feature.
We probably can have some rules like this: if a system output file from a classification task contains the feature of `probability`, then the processor will conduct calibration adaptively, (kinda similar to training set dependent features)

neubig:
At the moment, if an analysis is not applicable it returns `None`, so we could do a similar thing here.

pfliu-nlp:
Yes, I noticed that. But this also will result in potential bugs when deploying the web platforms, which I have debugged for a pretty long time.
It's a potential schema validation bug that will happen here: https://github.com/neulab/explainaboard_web/blob/c711cf8277c4f19d0c12e18b008b7ec2b8779d00/backend/src/impl/default_controllers_impl.py#L447

neubig:
I don't think returning `None` is necessarily a bad thing if we know it's expected behavior. But we could definitely discuss ways to rectify this if it's a problem.

odashi:
I just created #418 around the topic of `None`.

Basically I think `None` is not informative as it provides no fine-grained information, and users have no control in case of invalid operations. I prefer either:

- raising appropriate exceptions to inform appropriate information outside the process.
- Using the `Result` semantics (adopted in some other languages, such as [Rust](https://doc.rust-lang.org/std/result/enum.Result.html)). the result type have either the data or error as data.

odashi:
I think it is better that `None` is used to inform only *no information*. If there are some information that are useful to inform externally, it is good to adopt either way described above.

qjiang002:
Here are some draft ideas of implementing calibration analysis:
* When to perform calibration analysis: (1) classification tasks: text_classification, text_pair_classification, aspect_based_sentiment_classification, tabular_classification; (2) users provide both predicted labels and confidence values in the output file. Check that users provide confidence values in range [0, 1], not logits. 
[...]

PR #529, opened 2022-10-04
Implement calibration analysis

qjiang002:
Related issue: #417 
### Current settings of calibration analysis
* If there is Accuracy metric in the task's default metrics, automatically detect if there is confidence feature in the output json file and perform calibration analysis. It only supports json output file now. The tasks are text_classification, text_pair_classification, tabular_classification, aspect_based_sentiment_analysis
* The calibration analysis is formulated as a bucket analysis of the "confidence" feature. The buckets are fixed as 10 buckets evenly dividing the (0, 1) interval. This can lead to empty buckets and imbalanced buckets, but it allows comparison between different systems. The calibration errors ECE and MCE also depends on the bucketing, so keeping the bucketing fixed may reflect the performance change more accurately. 

Example input: `data/system_outputs/absa/absa-example-output-confidence.json`
```
"examples": [
    {
      "predicted_label": "positive",
      "confidence": 0.22101897026820283
    }
    ...
]
```
Example output: `data/reports/absa-confidence-report.json`
```
{
        "name": "confidence",
        "level": "example",
        "bucket_performances": [
          {
            "n_samples": 9,
            "bucket_samples": [
              7,
              11,
              18,
              55,
              62,
              66,
              83,
              91,
              94
            ],
            "performances": {
              "Accuracy": {
                "value": 1.0,
                "confidence_score_low": null,
                "confidence_score_high": null,
                "auxiliary_result": {
                  "confidence": 0.05506609222867546
                }
              }
            },
            "bucket_interval": [
              0.0,
              0.1
            ],
            "bucket_name": null
          },
          ...
        ],
        "cls_name": "BucketAnalysisResult",
        "auxiliary_performances": {
          "expected_calibration_error": 0.4588287500885495,
          "maximum_calibration_error": 0.9449339077713246
        }
      }
```

### Changes to existing classes
* Set a `skippable` property to FileLoader, FeatureType and Analysis classes. This is to automatically skip loading confidence feature or performing bucket analysis on confidence feature. 
* Put bucket average confidence value as `AuxiliaryMetricResult` in Accuracy's `Performance.auxiliary_result`. 
* Create `BucketAnalysisResult.auxiliary_performances` which has type `AuxiliaryAnalysisResult` to hold `CalibrationAnalysisResult`. `CalibrationAnalysisResult` holds the calibration errors ECE and MCE.
* Define confidence feature, calibration analysis, and json file loader fields in the classification tasks' loaders and processors.

### Testing
Test on tasks: text_classification, text_pair_classification, aspect_based_sentiment_analysis
Using both text output file and json output file with only predicted_label, and the report generated successfully.
Using json output file with predicted_label and confidence, the report generated successfully with calibration analysis (shown as confidence bucket analysis).


explainaboard/metrics/accuracy.py
```diff
@@ -17,6 +19,19 @@
...
+
+    This is an auxiliary result of Accuracy metric result.
+
+    Args:
+        confidence: The average confidence of a bucket of sampels.
+    """
+
+    confidence: float
+
+
```
odashi, line 34: There is already `Score` to store any float values. We basically don't need to implement `AuxiliaryMetricResult` anymore. You instead need to add additional results with `Score` in `evaluate_from_stats`.

https://github.com/neulab/ExplainaBoard/blob/00f1a6db749e6dd9b171d7192cbc7c0afa70fb75/explainaboard/metrics/metric.py#L43

explainaboard/analysis/analyses.py
```diff
@@ -234,26 +258,58 @@ class BucketAnalysis(Analysis):
     setting: Any = None  # For different bucket_methods, the settings are diverse
     sample_limit: int = 50
     cls_name: Optional[str] = None
+    auxiliary_analysis: Optional[bool] = False
+    skippable: Optional[bool] = False
```
odashi, line 262: The additional logic is almost independent from this analysis. I recommend to implement another Analysis class rather than extending this.

odashi (changes requested):

explainaboard/analysis/analyses.py
neubig: Just a note that we had an in-person meeting where we agreed that we'll follow this suggestion, thanks!

qjiang002: Thanks for the review! I'll make the change soon.

explainaboard/metrics/accuracy.py
qjiang002: Defining ConfidenceMetricResult inheriting AuxiliaryMetricResult is because the Performance class is defined as 
```
class Performance:
    value: float
    confidence_score_low: float | None = None
    confidence_score_high: float | None = None
    auxiliary_result: AuxiliaryMetricResult | None = None
```
If we add confidence value to the `MetricResult` in `evaluate_from_stats`, we cannot see it in `Performance` and still need to create a `AuxiliaryMetricResult`.

odashi: I see. We need the following approach:

- Keep `ConfidenceMetricResult` for now.
- `metric.evaluate_from_stats` should return `Score` with name "confidence" because AuxiliaryMetricResult is being deprecated. 
- Repack the "confidence" value in `Score` to `ConfidenceMetricResult` when `MetricResult` is converted to `Performance`

This is the same situation with "agreement" in `ExternalEval`, which is currently implemented as `Score`.
https://github.com/neulab/ExplainaBoard/blob/9ba998c125587bb5a9b421706b1ca50028af220b/explainaboard/metrics/external_eval.py#L175

qjiang002:
In the new commit, `CalibrationAnalysis` and `CalibrationAnalysisResult` are created in `analyses.py`. The default setting is to divide the [0, 1] interval into 10 buckets. In the classification task's default setting, it will automatically look for `confidence` feature and perform calibration analysis on `confidence`. It also supports custom calibration analysis as shown in `./data/system_outputs/absa/absa-example-output-custom-calibration-analysis.json`.

explainaboard/analysis/analyses.py
```diff
@@ -321,6 +333,253 @@ def perform(
...
+        bucket_func: Callable[..., list[AnalysisCaseCollection]] = getattr(
+            explainaboard.analysis.bucketing,
+            "fixed",
+        )
+        bucket_size = 1.0 / self.number
+        bucket_setting = [
+            (
+                (i * bucket_size, (i + 1) * bucket_size)
+                if i < self.number - 1
+                else (i * bucket_size, 1.0)
```
odashi, line 505: I think multiplication with a reciprocal causes certain amount of floating point errors.
Directly dividing the value would be better (and it is not so expensive in this case):
```suggestion
                (i / self.number, (i + 1) / self.number)
                if i < self.number - 1
                else (i / self.number, 1.0)
```

explainaboard/analysis/analyses.py
```diff
@@ -321,6 +333,253 @@ def perform(
...
+                ci_high: Optional[float] = None
+                auxiliary_result = None
+            else:
+                bucket_stats = metric_stat.filter(bucket_collection.samples)
+                metric_result = acc_metric.evaluate_from_stats(
+                    bucket_stats,
+                    confidence_alpha=confidence_alpha,
+                )
+                value = unwrap(metric_result.get_value(Score, "score")).value
+                confidence = unwrap(metric_result.get_value(Score, "confidence"))
```
odashi, line 541:
```suggestion
                confidence = metric_result.get_value(Score, "confidence")
```

explainaboard/analysis/analyses.py
```diff
@@ -321,6 +333,253 @@ def perform(
         )
 
 
+@final
+@dataclass
+class CalibrationAnalysisResult(AnalysisResult):
```
odashi, line 338: No unit tests about this class.

explainaboard/analysis/analyses.py
```diff
@@ -321,6 +333,253 @@ def perform(
...
+
+        texts.append(f"expected_calibration_error\t{self.expected_calibration_error}")
+        texts.append(f"maximum_calibration_error\t{self.maximum_calibration_error}")
+
+        return "\n".join(texts)
+
+
+@final
+@dataclass
+class CalibrationAnalysis(Analysis):
```
odashi, line 420: No unit tests about this class.

explainaboard/analysis/analyses.py
```diff
@@ -321,6 +333,253 @@ def perform(
...
+                    performances=performances,
+                    bucket_interval=bucket_collection.interval,
+                    bucket_name=bucket_collection.name,
+                )
+            )
+
+        (
+            expected_calibration_error,
+            maximum_calibration_error,
+        ) = self.perform_calibration_analysis(bucket_performances)
```
odashi, line 573: nit: It is OK to use shorter names since these variables are immediately used in the next statement.
```suggestion
        ece, mce = self.perform_calibration_analysis(bucket_performances)
```

explainaboard/analysis/analyses.py
```diff
@@ -321,6 +333,253 @@ def perform(
...
+    sample_limit: int = 50
+    cls_name: Optional[str] = None
+
+    def __post_init__(self):
+        """Set the class name."""
+        self.cls_name: str = self.__class__.__name__
+
+    AnalysisCaseType = TypeVar('AnalysisCaseType')
+
+    def perform_calibration_analysis(
```
odashi, line 444: This should be a private function.
```suggestion
    def _perform_calibration_analysis(
```

explainaboard/analysis/feature.py
```diff
@@ -39,19 +39,23 @@ def __init__(
         description: str | None = None,
         func: Callable[..., Any] | None = None,
         require_training_set: bool | None = None,
+        skippable: Optional[bool] = False,
```
odashi, line 42: - According to the description, the name seems to be "optional". "skippable" sounds somewhat confusing.
- We don't need to make this variable Optional because the default value `False` is given.
```suggestion
        optional: bool = False,
```

explainaboard/analysis/feature.py
```diff
@@ -39,19 +39,23 @@ def __init__(
...
     ) -> None:
         """Initializes FeatureType object.
 
         Args:
             description: Description of this feature.
             func: Function to calculate this feature from other features.
             require_training_set: Whether this feature relies on the training samples.
+            skippable: set it to True if this feature is optional in the output file
+                or it does not require default bucket analysis
```
odashi, line 51: We don't need to provide external contexts.
```suggestion
            skippable: set it to True if this feature is optional.
```

explainaboard/loaders/file_loader.py
```diff
@@ -62,6 +62,7 @@ class FileLoaderField:
     dtype: Optional[DType] = None
     strip_before_parsing: Optional[bool] = None
     parser: Optional[Callable[[str], str]] = None
+    skippable: Optional[bool] = False
```
odashi, line 65: Ditto
```suggestion
    optional: bool = False
```

explainaboard/metrics/accuracy.py
```diff
@@ -46,6 +66,86 @@ def calc_stats_from_data(
...
+    def _calc_metric_from_aggregate(
+        self,
+        agg_stats: np.ndarray[tuple[int], Any] | np.ndarray[tuple[int, int], Any],
+    ) -> np.ndarray[tuple[()], Any] | np.ndarray[tuple[int], Any]:
+        """Inner function of calc_metric_from_aggregate."""
+        if agg_stats.shape[-1] == 1:
+            return agg_stats.squeeze(-1)
+        return agg_stats
+
+    def calc_metric_from_aggregate(
```
odashi, line 78: This function is prohibited to be overridden.
https://github.com/neulab/ExplainaBoard/blob/f4c48e5f54418f0a5c4df10f93be43e80c986225/explainaboard/metrics/metric.py#L452-L453

explainaboard/metrics/accuracy.py
```diff
@@ -46,6 +66,86 @@ def calc_stats_from_data(
...
         )
 
+    def _calc_metric_from_aggregate(
+        self,
+        agg_stats: np.ndarray[tuple[int], Any] | np.ndarray[tuple[int, int], Any],
+    ) -> np.ndarray[tuple[()], Any] | np.ndarray[tuple[int], Any]:
+        """Inner function of calc_metric_from_aggregate."""
+        if agg_stats.shape[-1] == 1:
+            return agg_stats.squeeze(-1)
+        return agg_stats
```
odashi, line 76: I think you intended to return both the score and confidence from this function, but the caller of this function allows to return only the score. You need to implement another function to return the confidence scores.

explainaboard/metrics/accuracy.py
```diff
@@ -46,6 +66,86 @@ def calc_stats_from_data(
...
+
+        assert result.shape == result_shape, (
+            "BUG: invalid operation: "
+            f"{type(self).__name__}._calc_metric_from_aggregate(): "
+            f"Expected shape {result_shape}, but got {result.shape}."
+        )
+
+        return result
+
+    def evaluate_from_stats(
```
odashi, line 111: Don't override this function. Instead, directly implement statements to retrieve confidence scores in the original function (as just same as "score_ci" MetricValue).

odashi (commented):
It looks there is no unit tests for this change. Please add appropriate tests about changes you introduced.

explainaboard/analysis/feature.py
```diff
@@ -100,6 +104,12 @@ def require_training_set(self) -> bool:
         """Returns whether this feature requires training set or not."""
         return self._require_training_set
 
+    @final
+    @property
+    def skippable(self) -> bool:
+        """Returns the description of this feature."""
```
neubig, line 110:
```suggestion
        """Returns whether this feature is optional."""
```

explainaboard/analysis/feature.py
```diff
@@ -100,6 +104,12 @@ def require_training_set(self) -> bool:
...
+        return False if self._skippable is not True else True
```
neubig, line 111:
```suggestion
        return self._skippable
```

neubig (commented):
I read through and overall I think this looks good, but it'd be good to have @odashi 's comments reflected!

explainaboard/loaders/file_loader.py
qjiang002: `optional: bool = False` needs to be put before all `Optional` variables. Since some `FileLoaderField` definitions specify dtype without `dtype=`, it's better to put it at the end as `optional: Optional[bool] = None` similar to `strip_before_parsing`.

qjiang002:
Thank you for all your comments!
In the new commit, in Metric `evaluate_from_stats`, add `auxiliary_stats` which is used to calculate auxiliary metric results like confidence. This is because confidence values are not in the original metric stats which is calculated in the `get_overall_statistics`. A new `MetricStats` with confidence values are calculated in `CalibrationAnalysis`.

explainaboard/analysis/analyses.py
```diff
@@ -321,6 +333,264 @@ def perform(
...
+    feature: str
+    num_buckets: int = 10
+    sample_limit: int = 50
+    cls_name: Optional[str] = None
+
+    def __post_init__(self):
+        """Set the class name."""
+        self.cls_name: str = self.__class__.__name__
+        if self.num_buckets <= 0:
+            self.num_buckets = 10
```
odashi, line 451: Raise error rather than implicitly setting defaults. Implicit setting often causes unintended behavior.
```suggestion
        if self.num_buckets <= 0:
            raise ValueError(f"Invalid num_buckets: {num_buckets}")
```

explainaboard/loaders/file_loader.py
```diff
@@ -62,6 +62,7 @@ class FileLoaderField:
     dtype: Optional[DType] = None
     strip_before_parsing: Optional[bool] = None
     parser: Optional[Callable[[str], str]] = None
+    optional: Optional[bool] = None
```
odashi, line 65: Is there any reason to keep this value `Optional`? It looks this value is used only as `bool`: the value `None` looks meaningless.

explainaboard/metrics/metric.py
```diff
@@ -586,17 +586,25 @@ def calc_confidence_interval(
             high = int(num_iterations * (1.0 - confidence_alpha / 2.0))
             return float(samp_results[low]), float(samp_results[high])
 
+    def calc_metric_from_auxiliary_stats(
+        self, auxiliary_stats: MetricStats
+    ) -> np.ndarray[tuple[()], Any] | np.ndarray[tuple[int], Any]:
+        """Calculate the auxiliary result from auxiliary stats."""
+        pass
```
odashi, line 593:
```suggestion
        raise NotImplementedError
```

odashi (approved):
Some trivial comments:

explainaboard/loaders/file_loader.py
qjiang002: `optional: bool = False` needs to be put before all `Optional` variables. Since some FileLoaderField definitions specify dtype without `dtype=`, I think it can be put at the end similar to `strip_before_parsing`.

odashi: Requirements around the order of members is whether it has the default value or not. `optional: bool = False` correctly sets the default value to the member and it can be put after `Optional` members.
Note that `Optional[T] == T | None` only means that the member becomes either a value of `T` or `None`. Defaulting to `None` does not mean "unset".

Actually, all mypy/flake8/black passes the dataclass with defaults after Optional:
```shell
$ cat temp.py                                    
@dataclass
class Foo:
    a: int
    b: int | None = None
    c: int = 42

$ mypy temp.py && flake8 temp.py && black temp.py
Success: no issues found in 1 source file
All done! ✨ 🍰 ✨
1 file left unchanged.
```

qjiang002: I see. Thanks!

odashi (approved):

odashi:
@qjiang002 This error can be solved by using `get_value_or_none` instead.

https://github.com/neulab/ExplainaBoard/actions/runs/3212963391/jobs/5252276938#step:5:234

PR #531, opened 2022-10-05
skip docstring check for third party code

shuaichenchang:


shuaichenchang:
The PR adds all py code in the third-party folder in the exclude of the docstring check.

neubig (approved):
I approve, and in fact I think this is necessary since we generally will not be changing the third-party code.

PR #538, opened 2022-10-08
MetricResult.get_value_or_none

odashi:
This introduces following changes:

- `MetricResult.get_value`: returns `MetricValue` (originally `MetricValue | None`)
- Introduce `MetricResult.get_value_or_none`: returns `MetricValue | None`

In most cases `get_value` was used with `unwrap`, which increases the complexity of the code. This change mitigates some burden with keeping availability to obtain optional MetricValues.


pfliu-nlp (approved):
Yeah, this is a clever idea.

neubig (approved):
LGTM as well!

PR #539, opened 2022-10-08
Remove Performance

odashi:
This change removes `Performance` class and replaces all `Performance` instances to `MetricResult`. These classes holds the same information in the current repository.


explainaboard/analysis/performance.py
```diff
@@ -43,8 +45,13 @@ def dict_conv(cls, k: str, v: Any) -> Any:
         Returns:
             The value corresponding to the key
         """
+        serializer = PrimitiveSerializer()
+
         if k == 'performances':
```
pfliu-nlp: Should this be `results` instead of `performances`

pfliu-nlp: it seems that this has been fixed in: https://github.com/neulab/ExplainaBoard/pull/539/commits/34b0a6861adf054e9e7f68412ed104aa18626bb3

pfliu-nlp (approved):
Thanks, and it looks good to me.

neubig (approved):

odashi:
I think we need to merge #529 first. It involves overlapped changes and some migration is required.

odashi:
@neubig requesting review for recent cherrypicks

neubig (approved):
Looks good, I saw no issues.

PR #532, opened 2022-10-06
add third party code for text-to-sql evaluation

shuaichenchang:
Add third-party code for text-to-sql evaluation. The code is adapted from https://github.com/taoyds/test-suite-sql-eval

PR #545, opened 2022-10-09
Teach markdownlint-cli2 to ignore explainaboard/third_party

tetsuok:
Currently, the Markdown linter CLI, markdownlint-cli2 is not properly configured to ignore `explainaboard/third_party`. This is because we used a different type of the configuration file format `.markdownlint.yaml`, which is a configuration file for the library, `markdownlint`, not for the CLI. The configuration file doesn't support `ignores` key. We need to use a configuration file format for the CLI, `.markdownlint-cli2.yaml` which supports `ignores`key. This PR replaces the `.markdownlint.yaml` with `.markdownlint-cli2.yaml`, and fixes the syntax to control rules. (I'll apply similar change in explainaboard_web as well to avoid confusions caused by different config file formats across related repositories)


tetsuok:
@neubig This issue blocks #532. Would you review this PR? If this looks good, would you merge this PR for me, and then merge `main` into `third_party_code_for_text_to_sql_evaluation`?

odashi (approved):

odashi:
It looks this change is somewhat urgent: approved it on behalf of assigned reviewers.

tetsuok:
@odashi Thanks!

PR #548, opened 2022-10-09
fix bug: wrong error type

pfliu-nlp:
This PR fixed a wrong specification of error type: `ValueError` -> `FileNotFoundError`

### DIFF
--- a/explainaboard/processors/processor.py
+++ b/explainaboard/processors/processor.py
@@ -194,7 +194,7 @@ class Processor(metaclass=abc.ABCMeta):
                         output_data=None,
                     )
                     dataset = loader.load()
-                except ValueError as e:
+                except FileNotFoundError as e:
                     get_logger().warning(
                         f"{sys_info.dataset_name} could not be loaded by DataLab so"
                         " no training set dependent features will be supported by"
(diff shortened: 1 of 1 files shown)


neubig (approved):

explainaboard/processors/processor.py
```diff
@@ -194,12 +194,16 @@ def _gen_external_stats(self, sys_info: SysOutputInfo) -> Any:
                         output_data=None,
                     )
                     dataset = loader.load()
-                except ValueError as e:
+                except FileNotFoundError as e:
```
odashi, line 197: This is weird: `FileNotFoundError` is corresponding to the `ENOENT` system errno by design and using it for any other purpose confuses the entire system.

pfliu-nlp: Hi, @odashi.  Here `FileNotFoundError` is used to capture the error raised by DataLab when invalid dataset names are specified by users.  DataLab adopt huggingface dataset's strategy for dataset loading process, specifically,
* it will first download all datasets' dataloader script files [here]FileNotFoundError(https://github.com/ExpressAI/DataLab/tree/main/datasets)
* then judge if the user-specified dataset names could be found from the above folder,
* if not, `FileNotFoundError` will be raised.

odashi: @pfliu-nlp I meant you need to revise the strategy in Datalab. It is dangerous to use built-in errors for other meanings.

PR #561, opened 2022-10-14
Implement Serializable interface in BucketPerformance

odashi:
# Overview

This change introduces `Serializable` interface to `BucketPerformance` and removes `from_dict` functionality.

# Details

`BucketPerformance` now has only `Serializable` data and can also implement `Serializable` for itself.
This change involves following implementations:

- `Serializable` interface of `BucketPerformance` with fine-grained type checking.
- Add `MetricResult.__eq__` to make `BucketPerformance` comparable.

# References

- #446

# Blocked by

- #539


explainaboard/analysis/performance.py
```diff
@@ -30,35 +33,55 @@ class BucketPerformance:
...
-        Args:
-            k: the parameter name
-            v: the parameter's value
+    def __post_init__(self) -> None:
+        """Validates member values."""
+        has_interval = self.bucket_interval is not None
+        has_name = self.bucket_name is not None
+        if has_interval == has_name:
+            raise ValueError(
+                "Only either `bucket_interavl` or `bucket_name` must has a value. "
```
neubig, line 42:
```suggestion
                "Either `bucket_interval` or `bucket_name` must have a value, but not both. "
```

explainaboard/analysis/performance.py
```diff
@@ -30,35 +33,55 @@ class BucketPerformance:
...
+                f"{self.bucket_interval=}, {self.bucket_name=}"
```
neubig, line 43: Just a comment, this is a cool feature that I didn't know of! https://stackoverflow.com/questions/59661904/what-does-equal-do-in-f-strings-inside-the-expression-curly-brackets

neubig (approved):

PR #550, opened 2022-10-09
Add temporary fix to MetaEvalNLGCITest to fix flakiness

tetsuok:
This PR adds a workaround to fix the flakiness of `MetaEvalNLGCITest` as discussed in #537. The code change is cherry-picked from #532 (including suggestions by @neubig) because other pull requests get affected by the flakiness and because #532 is still in review.

PR #565, opened 2022-10-15
Add seed argument to Metric

### LINKED ISSUE #537: Potential issue with spearman R bootstrapping
We observed the following test failure when i[ntegrating another PR](https://github.com/neulab/ExplainaBoard/pull/532):
```
======================================================================
FAIL: test_sample_level_spearmanr_bootstrap (integration_tests.meta_eval_wmt_da_test.MetaEvalNLGCITest)
----------------------------------------------------------------------
  File "/home/runner/work/ExplainaBoard/ExplainaBoard/integration_tests/meta_eval_wmt_da_test.py", line 191, in test_sample_level_spearmanr_bootstrap
    self.assertAlmostEqual(ci[0], 0.6488, 2)
AssertionError: 0.7325904563487001 != 0.6488 within 2 places (0.08379045634870008 difference)

----------------------------------------------------------------------
```

We are not sure whether this is an issue with the test or the underlying code, but as a temporary measure we reduced the sensitivity of the test. We should go back and check to make sure whether this is just due to bootstrapping variance or whether it's due to a bug in the test itself.

tetsuok:
# Overview

This PR adds `seed` argument to `Metric` class to fix #537.

# Details

`Metric`'s `calc_confidence_interval` method uses random generator for bootstrapping, but we have no way to provide a seed to the method to make the behavior reproducible. To address this issue, adds `seed` argument to `Metric` class, as
described in https://github.com/neulab/ExplainaBoard/issues/537#issuecomment-1274340748.

As per this change, reverted a hot fix in #550. This is to use almost equal assertions to assert confidence intervals computed by bootstrapping.

# References

- #537

# Blocked by

- NA

### DIFF
--- a/explainaboard/metrics/metric.py
+++ b/explainaboard/metrics/metric.py
@@ -396,7 +396,9 @@ class Metric(metaclass=abc.ABCMeta):
 
     config: MetricConfig
 
-    def __init__(self, config: MetricConfig, seed: np.random.SeedSequence | None):
+    def __init__(
+        self, config: MetricConfig, seed: np.random.SeedSequence | None = None
+    ):
         """Initialize the metric.
 
         Args:
--- a/explainaboard/metrics/metric.py
+++ b/explainaboard/metrics/metric.py
@@ -431,7 +431,19 @@ class Metric(metaclass=abc.ABCMeta):
     def get_seed(self) -> np.random.SeedSequence:
         """Gets a numpy SeedSequence.
 
-        See Metric for the usage.
+        Returned SeedSequence can be used to construct other classes by this method.
+        See also the `Metric` docstring.
+
+        Example:
+            class Routine:
+                def __init__(self, metric: Metric):
+                    self._metric = metric
+                    seed = self._metric.get_seed().spawn(1)[0]
+                    # Supposes that a class, InnerRoutine holds its own SeedSequence.
+                    self._inner_routine = InnerRoutine(seed=seed)
+
+                def do_something(self):
+                    self._inner_routine.do_something_else()
 
         Returns:
             A numpy SeedSequence.
--- a/explainaboard/metrics/metric.py
+++ b/explainaboard/metrics/metric.py
@@ -380,7 +380,7 @@ class Metric(metaclass=abc.ABCMeta):
 
         Args:
             config: The configuration for the metric
-            seed: a user-provided seed to initialize a SeedSequence.
+            seed: A user-provided seed to initialize a SeedSequence.
         """
         self.config = config
         self._seed = np.random.SeedSequence(seed)
--- a/explainaboard/metrics/metric.py
+++ b/explainaboard/metrics/metric.py
@@ -407,14 +407,13 @@ class Metric(metaclass=abc.ABCMeta):
 
     @final
     def get_seed(self) -> np.random.SeedSequence:
-        """Gets a SeedSequence.
+        """Gets a numpy SeedSequence.
 
         When a subclass needs to construct a random number generator, initialize the
-        generator with a spawned `numpy.random.SeedSequence` object from the
-        `SeedSequence` object returned by this method.
+        generator by spawned numpy SeedSequence from the returned SeedSequence object.
 
         Returns:
-            A SeedSequence.
+            A numpy SeedSequence.
         """
         return self._seed
 
--- a/explainaboard/metrics/metric.py
+++ b/explainaboard/metrics/metric.py
@@ -375,12 +375,15 @@ class Metric(metaclass=abc.ABCMeta):
 
     config: MetricConfig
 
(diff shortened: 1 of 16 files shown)


explainaboard/metrics/metric.py
```diff
@@ -371,16 +371,39 @@ def get_batch_data(self) -> np.ndarray[tuple[int, int, int], Any]:
 
 
 class Metric(metaclass=abc.ABCMeta):
-    """A class representing an evaluation metric."""
+    """A class representing an evaluation metric.
+
+    When a subclass needs to construct a random number generator, initialize the random
+    generator as follows: Invoke the `get_seed()` method to get a numpy SeedSequence,
+    spawn the SeedSequence with the SeedSequence's `spawn` method, and then pass the
+    spawned SeedSequence to the random generator's constructor.
```
odashi, line 379: Better to organize this section using enumeration.

explainaboard/metrics/metric.py
```diff
@@ -371,16 +371,39 @@ def get_batch_data(self) -> np.ndarray[tuple[int, int, int], Any]:
...
+
+        config = MetricConfig(...)
+        metric = FooMetric(config, seed=12345)
+        metric.foo()
+    """
 
     config: MetricConfig
 
-    def __init__(self, config: MetricConfig):
+    def __init__(self, config: MetricConfig, seed: int | None = None):
```
odashi, line 397: Give SeedSequence rather than int so that this class can be constructed by other classes that may hold their own SeedSequence:
```python
class Routine:
  def __init__(self, seed: np.SeedSequence | None):
    self._seed = seed if seed is not None else np.SeedSequence()
    self._inner_routine = InnerRoutine(seed=self._seed.spawn())

  def do_something(self):
    self._inner_routine.do_something_else()
```

explainaboard/metrics/metric.py
```diff
@@ -402,6 +425,17 @@ def calc_stats_from_data(
         """
         ...
 
+    @final
+    def get_seed(self) -> np.random.SeedSequence:
+        """Gets a numpy SeedSequence.
+
+        See Metric for the usage.
```
odashi, line 432: Details should be written here as well.

explainaboard/metrics/metric.py
tetsuok: Changed to take `np.random.SeedSequence`.

explainaboard/metrics/metric.py
tetsuok: Done.

explainaboard/metrics/metric.py
tetsuok: Added more explanations to the docstring.

explainaboard/metrics/metric.py
```diff
@@ -371,16 +371,41 @@ def get_batch_data(self) -> np.ndarray[tuple[int, int, int], Any]:
...
+
+        config = MetricConfig(...)
+        metric = FooMetric(config, seed=np.random.SeedSequence(12345))
+        metric.foo()
+    """
 
     config: MetricConfig
 
-    def __init__(self, config: MetricConfig):
+    def __init__(self, config: MetricConfig, seed: np.random.SeedSequence | None):
```
odashi, line 399: It looks the default value `None` has been removed by the new commit. Is there any intent of this?

explainaboard/metrics/metric.py
```diff
@@ -429,7 +431,19 @@ def calc_stats_from_data(
...
+        Example:
+            class Routine:
+                def __init__(self, metric: Metric):
+                    self._metric = metric
+                    seed = self._metric.get_seed().spawn(1)[0]
+                    # Supposes that a class, InnerRoutine holds its own SeedSequence.
+                    self._inner_routine = InnerRoutine(seed=seed)
+
+                def do_something(self):
+                    self._inner_routine.do_something_else()
```
odashi, line 446: I think this example is weird: since metric is created outside and modified many times according to the invocation order of the program, it does not guarantee the reproducibility of the behavior.
What I meant in the previous comment is taking the exact seed (or None) by the class that is required to initialize some inner class. In this case, the Routine takes SeedSequence rather than Metric, and initializes Metric by itself.

explainaboard/metrics/metric.py
tetsuok: Ah, I was confused with the previous sample code which doesn't have None as the default value. Will bring back the default value of None.

explainaboard/metrics/metric.py
tetsuok: I see. I understand your point. Will update the sample code.

explainaboard/metrics/metric.py
tetsuok: Reverted the change which removed `None`.

odashi (approved):

PR #569, opened 2022-10-18
fix wrong specification of column name

pfliu-nlp:
# Overview
For the `datalab` data loader of `absa` task, there are specifications of column names.

# Details

# References

# Blocked by


neubig (approved):
I approve, but note that we should start moving data for the integration tests out of the EB repo and into the inspired-public GCP bucket. Maybe we can do that for the whole `integration_tests/artifacts` directory soon.

PR #570, opened 2022-10-18
Lint missing '__future__ import annotations' via pre-commit and CI

tetsuok:
# Overview

This PR introduces a flake8 plugin, [flake8-future-import](https://pypi.org/project/flake8-future-import/) to automatically check missing `__future__ import annotations`.

# Details

Currently, some code requires `__future__ import annotations` to use type annotation (e.g., [PEP 604](https://peps.python.org/pep-0604/) available in Python 3.10) across Python 3.8, 3.9, and 3.10. It's more consistent to add the import for every Python file except third party, but it's easy to forget to add it without CI. That's where flake8-future-import comes in.

flake8-future-import enables flake8 to check the import automatically. This PR adds a new pre-commit hook dedicated to the plugin to exclude third-party code from the check.  I added error codes to the config file to silence flake8 so that flake8 doesn't check features other than `annotations`, because the plugin is designed to lint all the features by default.
Note that this PR bumps flake8 version to use the latest version of the plugin.

# References

# Blocked by

- NA

### DIFF
--- a/explainaboard/analysis/__init__.py
+++ b/explainaboard/analysis/__init__.py
@@ -1,1 +1,3 @@
+# flake8: noqa: FI18
+
 """This directory contains code implementing various analyses."""
--- a/explainaboard/analysis/__init__.py
+++ b/explainaboard/analysis/__init__.py
@@ -1,3 +1,3 @@
-# flake8: noqa: FI18
-
 """This directory contains code implementing various analyses."""
+
+from __future__ import annotations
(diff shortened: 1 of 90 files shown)


explainaboard/analysis/__init__.py
```diff
@@ -1 +1,3 @@
+# flake8: noqa: FI18
```
odashi, line 1: I think importing is better even in `__init__.py`.

tetsuok: I'll add the import statement to `__init__.py` as I don't have a strong opinion.

tetsuok: Added.

neubig (approved):

PR #571, opened 2022-10-18
Make Result/AnalysisResult Serializable.

odashi:
# Overview

This change introduces `Serializable` interface to `Result` and `AnalysisResult` with a reorganizing the internal structure.

# Details

`AnalysisResult` is originally an extensible dataclass. This design is problematic because all subclasses have to implement serialization of base members.

This change introduces `AnalysisDetails` abstract marker class, which is inherited by specific classes representing details of analyses. `AnalysisResult` holds `details: AnalysisDetails` member to represent its extension. This design allows to write serialization of each member only once.

(Also this design avoids some potential bugs around initializers of extensible dataclasses: if the base class changes the number of members, the arrangement of positional arguments of the subclass initializer also affects, causing potential failures).

## other changes:

- Add validation to `BucketAnalysisDetails` and `CalibrationAnalysisDetails` to check if it has at least 1 `BucketPerformance`.
- Reorganized failure tests in `analyses_test.py` to check only the targeted function.

# References

- #446

# Blocked by

- NA


odashi:
I noticed that there is no tests for the change. fixing it--

neubig (approved):
Looks good, thanks!

There is probably one typing error that needs to be fixed.

odashi:
Merging main, since #570 added a useful CI to figure out this error.

PR #578, opened 2022-10-20
remove unused functions in sysoutputinfo

odashi:
# Overview

Removes several functions in `SysOutputInfo`, which are not referenced from anywhere.

# Details

# References

# Blocked by

- NA


neubig (approved):
I approve, but it'd be nice to have a check from @pfliu-nlp

pfliu-nlp (approved):

PR #579, opened 2022-10-20
Serializable SysOutputInfo

### LINKED ISSUE #446: Centralized type registry

odashi:
# Overview

Makes SysOutputInfo serializable.
Fixes #446 

# Details

By applying this change, we can serialize almost everything that the ExplainaBoard generates.
Following code is also deleted because it is no longer necessary:

- Methods in SysOutputInfo to convert it to dict.
- `explainaboard/serialization/legacy.py`

This PR also fixes several typing bugs around SysOutputInfo that does not match with actual data.

# References

- #446

# Blocked by

- NA


odashi:
@neubig @pfliu-nlp It looks there is some failing tests (due to assertion around t-test), which is unexpected and may be a potential bug around the code, but it is somewhat hard to figure out the reason. Could you take a look at the errors?

neubig (approved):
Looks great, thanks!

odashi:
@neubig @pfliu-nlp Fixed bugs. Please take a look at the changes.

pfliu-nlp (approved):

Issue #518, opened 2022-10-02
AnalysisLevel does not own distinctive information


odashi:
Currently, every `Processor` distinguishes the process according to `AnalysisLevel`. This struct has name, features, and metric configs. Features and metric configs are specified by the processor itself, and they are only meaningful if the level is used with the same Processor. This means that the owner of these information is `Processor`
Thus, `AnalysisLevel` is eventually only the key of these values, and not necessary to be formed as a struct. I think it is suitable to (1) remove this struct and (2) return just `str` as the key of the analysis names.

RFC: @neubig

neubig:
I think that's a reasonable alternative design and may be simpler.

My only concern (which I haven't thought about very carefully yet) is whether changing to this design would limit us in our ability to make features more customizable, which is one of our high priority tasks.

odashi:
I think it is enough to provide a functionality to retrieve a list of features corresponding to given analysis level name.
```python
class MyProcessor(Processor):
  def get_feature_set(name: str) -> List[Feature]:
    if name == "foo":
      return [...]
    elif name == "bar":
      return [...]
```

The returned features can be modified outside the processor.

odashi:
I also think that the current AnalysisLevel can become an attribute of the Processor at initialization, or we can implement separate Processors for different AnalysisLevels, to avoid aditional control flow around AnalysisLevels.

neubig:
The solution in the first of the two above comments seems reasonable to me for now! The latter ones seem like bigger changes that we'd have to think about more carefully

PR #581, opened 2022-10-24
Fix Processor._customize_analyses

odashi:
# Overview

This PR restores the previous behavior in `Processor._customize_analyses`.

# Details

`Processor._customize_analyses` was changed resently to "add" given metrics to default metrics, but the original behavior was "replace" it with the given metrics. This change restores the original behavior.
This change also introduces `MetricConfig.replace_languages` and `AnalysisLevel.replace_metric_configs` functions, which creates a new object by copying self with replacing several members.

# References

# Blocked by

- NA

### DIFF
--- a/explainaboard/processors/processor.py
+++ b/explainaboard/processors/processor.py
@@ -250,26 +250,35 @@ class Processor(metaclass=abc.ABCMeta):
 
         Args:
             custom_features: the features to customize
-            metric_configs: additional metric configurations. Keys are analysis level
-                name and metric name.
+            metric_configs: MetricConfgs to replace.
+                If `metric_configs[analysis_level_name]` has a dict, it is used instead
+                of the default MetricConfigs associated to `analysis_level_name`.
             custom_analyses: the analyses to customize
 
         Returns:
             Customized analyses.
         """
-        analysis_levels = self.default_analysis_levels()
-        analyses = self.default_analyses()
-        for level in analysis_levels:
-            for name, config in metric_configs.get(level.name, {}).items():
-                level.metric_configs[name] = config
-            for config in level.metric_configs.values():
-                config.source_language = sys_info.source_language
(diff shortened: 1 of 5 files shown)


neubig:
Looks like there is a type-related error on Python 3.8, but I can take a look after that.

odashi:
@neubig It was resolved.

explainaboard/processors/processor.py
```diff
@@ -516,12 +525,19 @@ def get_overall_statistics(
         custom_features: dict = metadata.get('custom_features', {})
         custom_analyses: list = metadata.get('custom_analyses', [])
 
-        metric_configs: dict[str, dict[str, MetricConfig]] = {
-            "example": metadata.get('metric_configs', {})
-        }
+        metric_configs = metadata.get("metric_configs")
+        if metric_configs is not None:
+            metric_configs_dict = {
+                "example": {
```
lyuyangh, line 531: Thanks for fixing this! Just to clarify, it looks like `metric_configs` can only be used to replace `example` level metrics. Metrics for other levels (e.g. span) will always be the default ones. Is this intentional?

odashi: This keeps the original behavior, since metric_configs in metadata can only represent one analysis level. If we need to modify other levels, we need some additional syntax on metadata.

(I think this behavior could be resolved if we used analysis levels to only lookup default features and metrics. This is somewhat off-topic, and would be good to discuss on #518)

lyuyangh: Thanks! I think the current behavior is kind of unusual because users can override the `example` level but not the other ones. But it makes sense to leave it as it is for now because metadata doesn't support additional analysis levels.

neubig (approved):

pfliu-nlp (approved):

PR #591, opened 2022-11-06
Add text-to-SQL loader

shuaichenchang:
# Overall
The second step of adding the text-to-SQL task in https://github.com/neulab/ExplainaBoard/pull/580 
# Details
Standard input files in the text-to-SQL task use a JSON file containing the database id, question, and gold query in each example.  Standard output files are text files, where each line has a predicted query.

### DIFF
--- a/explainaboard/loaders/loader_factory.py
+++ b/explainaboard/loaders/loader_factory.py
@@ -31,6 +31,7 @@ from explainaboard.loaders.tabular_classification import TabularClassificationLo
 from explainaboard.loaders.tabular_regression import TabularRegressionLoader
 from explainaboard.loaders.text_classification import TextClassificationLoader
 from explainaboard.loaders.text_pair_classification import TextPairClassificationLoader
+from explainaboard.loaders.text_to_sql import TextToSQLLoader
 
 _LOADERS: dict[TaskType, type[Loader]] = {
     TaskType.argument_pair_extraction: ArgumentPairExtractionLoader,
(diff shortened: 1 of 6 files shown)


explainaboard/loaders/loader_factory.py
```diff
@@ -31,6 +31,7 @@
 from explainaboard.loaders.tabular_regression import TabularRegressionLoader
 from explainaboard.loaders.text_classification import TextClassificationLoader
 from explainaboard.loaders.text_pair_classification import TextPairClassificationLoader
+from explainaboard.loaders.text_to_sql import TextToSQLLoader
```
odashi, line 34: I think this always requires additional installation. I think we could no longer separate this loader from others, and the following installation flag needs to be integrated into the default.

https://github.com/neulab/ExplainaBoard/blob/0a9b3bcefefcb96751fa3d663d24fff2df9dd250/setup.cfg#L39

shuaichenchang: Thank you for the review! Do you want me to create a separate PR to merge the required installations into default?

odashi: No, it would be reasonable to mix it into this request I think.

shuaichenchang: Mixed it here. Thank you!

neubig (approved):

odashi (approved):

PR #593, opened 2022-11-09
add text_to_sql_processor

shuaichenchang:
# Overall
The third step of adding the text-to-SQL task in https://github.com/neulab/ExplainaBoard/pull/580 
Next Step: Add task document in a separate PR. 

# Details
Added the text-to-SQL processor and the integration test for the task loader, processor, and metric, along with artifact input/output files


explainaboard/processors/text_to_sql.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def _get_true_label(self, data_point):
+        """Override the parent class.
+
+        Return a pair of a SQL query and the database it corresponds to.
+        See processor._get_true_label for more details.
+        """
+        return [data_point["true_sql"], data_point["db_id"]]
+
+    def _get_predicted_label(self, data_point):
```
odashi, line 101:
```suggestion
    def _get_predicted_label(self, data_point: dict):
```

explainaboard/processors/text_to_sql.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Return a pair of a SQL query and the database it corresponds to.
+        See processor._get_predicted_label for more details.
+        """
+        return [
+            data_point["predicted_sql"].split("\t")[0],
+            data_point["predicted_sql"].split("\t")[1],
+        ]
+
+    def _statistics_func(self, samples: Iterable[Any], sys_info: SysOutputInfo):
```
odashi, line 112:
```suggestion
    def _statistics_func(self, samples: Iterable[Any], sys_info: SysOutputInfo) -> Any:
```

integration_tests/artifacts/text_to_sql/preds.sql
```diff
@@ -0,0 +1,3 @@
+select distinct country from singer where age > 20	concert_singer
+select distinct country from singer where age > 25	concert_singer
+select distinct country from singer where age = 20	concert_singer
```
odashi, line 3: Add newline at the EOF

integration_tests/artifacts/text_to_sql/questions.json
```diff
@@ -0,0 +1,17 @@
...
+        "db_id": "concert_singer",
+        "query": "select distinct country from singer where age > 20",
+        "question": "List the countries which have singers who are over 20."
+    },
+    {
+        "db_id": "concert_singer",
+        "query": "select distinct country from singer where age > 20",
+        "question": "What countries have singers who are over 20?"
+    }
+]
```
odashi, line 17: ditto

integration_tests/text_to_sql_test.py
```diff
@@ -0,0 +1,66 @@
...
+            self.json_dataset,
+            self.txt_output,
+            dataset_file_type=FileType.json,
+            output_file_type=FileType.text,
+        )
+        data = loader.load()
+        processor = get_processor_class(TaskType.text_to_sql)()
+        sys_info = processor.process(metadata, data, skip_failed_analyses=True)
+        self.assertGreater(len(sys_info.results.analyses), 0)
+        self.assertGreater(len(sys_info.results.overall), 0)
```
odashi, line 54: I think these tests can be replaced by exact matching if the dataset  is fixed.

integration_tests/artifacts/text_to_sql/preds.sql
shuaichenchang: done

integration_tests/text_to_sql_test.py
shuaichenchang: If I understand right, these mean the numbers of features/analyses (i.e. question_length and sql_length). So is it better to leave it loose if we want to add new features/analyses in the future?

shuaichenchang: But it also makes sense to modify these numbers whenever we want to add/remove features. I will change it according to your suggestion now.

odashi: General suggestion: if we could tighten the accepted behavior of unit tests, we should do so as much as possible.

In this case, specifying the number of objects can capture the unintended behavior that accidentally adds extra analyses.

shuaichenchang: That makes sense. Have made it specific. Thank you!

explainaboard/processors/text_to_sql.py
```diff
@@ -0,0 +1,117 @@
...
+                description="input question length in tokens",
+                func=lambda info, x, c: count_tokens(info, x["question"]),
+            ),
+            "sql_length": feature.Value(
+                dtype=feature.DataType.FLOAT,
+                description="context length in tokens",
+                func=lambda info, x, c: count_tokens(info, x["true_sql"]),
+            ),
+        }
+
```
neubig, line 61: I think it might be a good idea to add some features that would make the evaluation a little bit more insightful. For example, could we add something like the following (where we fill in a few more things for `SQL_OPERATIONS`)?

```suggestion

        SQL_OPERATIONS = ["select", "where", "distinct", ...]
        for op_name in SQL_OPERATIONS:
            features[f"contains_{op_name}"] = feature.Value(
                dtype=feature.DataType.STRING,
                description=f"whether the SQL contains '{op_name}'",
                func=lambda info, x, c: "yes" if f"{op_name} " in x["true_sql"]) else "no",
            )

```

explainaboard/processors/text_to_sql.py
```diff
@@ -0,0 +1,117 @@
...
+            AnalysisLevel(
+                name="example",
+                features=features,
+                metric_configs=self.default_metrics(),
+            )
+        ]
+
+    def default_analyses(self) -> list[Analysis]:
+        """See Processor.default_analyses."""
+        return self.continuous_feature_analyses()
```
neubig, line 72: If my above suggestion is reflected, this will need to be modified to consider the operations.

integration_tests/artifacts/text_to_sql/questions.json
```diff
@@ -0,0 +1,17 @@
+[
+    {
+        "db_id": "concert_singer",
+        "query": "select distinct country from singer where age > 20",
```
neubig, line 4: It might be nice to have tests that try a variety of DB queries, particularly if my suggestion for features over SQL operations is reflected.

neubig (commented):
Thanks @shuaichenchang ! This is great, but I think the features that we're calculating are rather shallow right now. I have one concrete suggestion that is probably not hard to implement but I think would greatly improve the utility of the analysis.

explainaboard/processors/text_to_sql.py
shuaichenchang: @neubig That's a great suggestion. I was planning to go through all steps of adding a task first and then come back to add more features. But the analysis regarding to SQL components sounds not hard to do in this repo. My goal is to add a dataset and its evaluation that we recently proposed for text-to-SQL model robustness.

shuaichenchang: @odashi The SQL operations/columns are case-insensitive during the evaluation. Here I can make the entire string lowercase to detect the SQL operations easier.

odashi: @shuaichenchang I'm not familiar with this task, but in general the entire SQL (not only the command, but also data) is case sensitive. If the safety is guaranteed by the task, I have no reason to apply lowercasing, but we also need to put some comments about explaining the safety.

shuaichenchang: Added a comment in the analysis. "Note: We use SQLite to evaluate SQLs, where SQL keywords and table/column names are case-insensitive." I double-checked that the data (string value in SQL and database cell value) are case sensitive but keywords and schemas are not in SQLite.

explainaboard/processors/text_to_sql.py
shuaichenchang: done

integration_tests/artifacts/text_to_sql/questions.json
shuaichenchang: Added another example with different keywords to test.

odashi (approved):

explainaboard/processors/text_to_sql_test.py
```diff
@@ -0,0 +1,14 @@
...
+from __future__ import annotations
+
+import unittest
+
+from explainaboard.constants import TaskType
+from explainaboard.processors.processor_factory import get_processor_class
+from explainaboard.processors.text_to_sql import TextToSQLProcessor
+
+
+class QAOpenDomainProcessorTest(unittest.TestCase):
```
neubig, line 12:
```suggestion
class TextToSQLProcessorTest(unittest.TestCase):
```

neubig (approved):
Looks great, thanks for making the change!

I noticed one typo as I was reading through a last time (sorry for not noticing earlier), but once that's fixed we can merge.

explainaboard/processors/text_to_sql_test.py
shuaichenchang: Oops. Thank you for pointing this out. I copied the function from the QA file but forgot the change the name. Have corrected it in the new commit.

Issue #575, opened 2022-10-20
Add system metadata class


odashi:
`Processor.process()` takes `metadata`, which is used to directly initialize `SysOutputInfo`. However, these are essentially different data (especially, "metadata" $\subset$ `SysOutputInfo`, but not $=$) and the current implementation makes some confusion around this:

The most significant abuse around this behavior is that `FileLoaderMetadata` is implicitly converted into `SysOutputInfo`. This shouldn't work unless explicit conversion: https://github.com/neulab/ExplainaBoard/blob/4cec0a01cbe2617e9a67a440be25ee4252f792b2/integration_tests/ner_test.py#L148-L154

To this end, we need:
- A struct defining the system metadata.
- Either:
  - A conversion method between system metadata and `FileLoaderReturn`/`SysOutputInfo`
  - Include system metadata as a direct member of `FileLoaderReturn`/`SysOutputInfo`

neubig:
Sounds like a good idea, I agree

PR #601, opened 2022-12-14
fix custom features/analyses loader

### LINKED ISSUE #596: `custom_features` and `custom_analyses` are not deserialized
https://github.com/neulab/ExplainaBoard/blob/e111afc94d01cc5ac62e7635bb2122b32b69ac4a/explainaboard/loaders/file_loader.py#L138-L151

`data["custom_analyses"]` is currently a `list[dict]`. It should have been deserialized like `custom_features` because `FileLoaderMetadata` expects a `list[Analysis]`.

Also, it looks like dataset `custom_features` & `custom_analyses` are applied twice. Once with deserialization and the second time without.

https://github.com/neulab/ExplainaBoard/blob/e111afc94d01cc5ac62e7635bb2122b32b69ac4a/explainaboard/loaders/file_loader.py#L647-L666
https://github.com/neulab/ExplainaBoard/blob/e111afc94d01cc5ac62e7635bb2122b32b69ac4a/explainaboard/loaders/file_loader.py#L690-L701

lyuyangh:
# Overview

We allow `custom_features` and `custom_analyses` to be defined for each system. This information is provided as a `dict` and it may come from the dataset or the system output file provided by the user. In some cases, it is not serialized so there is a type mismatch. This PR fixes the issue. Please see #596 for details.

# Details

# References

- closes #596 

# Blocked by

- NA

### DIFF
--- a/explainaboard/loaders/file_loader.py
+++ b/explainaboard/loaders/file_loader.py
@@ -135,20 +135,21 @@ class FileLoaderMetadata:
             source_language = target_language = data.get("language")
         custom_features: dict[str, dict[str, FeatureType]] = {}
         custom_analyses: list[Analysis] = []
+        serializer = PrimitiveSerializer()
         if "custom_features" in data:
-            ft_serializer = PrimitiveSerializer()
             custom_features = {
                 k1: {
                     # See https://github.com/python/mypy/issues/4717
-                    k2: narrow(
-                        FeatureType, ft_serializer.deserialize(v2)  # type: ignore
-                    )
+                    k2: narrow(FeatureType, serializer.deserialize(v2))  # type: ignore
                     for k2, v2 in v1.items()
                 }
                 for k1, v1 in data["custom_features"].items()
             }
         if "custom_analyses" in data:
-            custom_analyses = data["custom_analyses"]
+            custom_analyses = [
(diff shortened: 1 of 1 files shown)


odashi (approved):

odashi:
Test is failing for some reason.

lyuyangh:
Thanks, @odashi! I fixed the tests. 

A couple of pointers for what I changed:

1. `Processor.process()` takes deserialized `custom_features` and `custom_analyses` (according to the README and the integration tests) so the deserialize logic in `Processor._custom_analyses()` isn't necessary. That's why I removed it.
2. `MachineTranslationTest.test_custom_features()` was failing because `dataclasses.asdict()` serializes the two fields to `dict` but, as I mentioned above, `Processor.process()` expects deserialized objects. I am not sure which fields are actually being used by the processor so I followed the current implementation and included all the fields in `FileLoaderMetadata` when creating `metadata`. When #575 is implemented, this code probably needs to be updated.

odashi:
Yeah, It looks the original code was weird, because {de}serialization is the IO feature and shouldn't be happened in any inner processes (`Processor` here). The change looks good.

Also, in general we shouldn't use `dataclasses.asdict` due to its too generic behavior: it expands all nested dataclasses in the object, that possibly brings several breakage into the code.

PR #603, opened 2022-12-21
Deal with the case when correlation is NaN for meta evaluation task

pfliu-nlp:
# Overview

Currently, the correlation value of the meta-evaluation task could be NaN, which will break downstream applications. This PR fix this by setting them to 0.

### DIFF
--- a/explainaboard/metrics/meta_evaluation.py
+++ b/explainaboard/metrics/meta_evaluation.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-import math
 from typing import Any, Optional, Union
 
 import numpy as np
@@ -122,14 +121,14 @@ class CorrelationNLG(Metric):
         config = narrow(CorrelationNLGConfig, self.config)
         corr_func = config.get_correlation_func(config.correlation_type)
         if config.group_by == "dataset":
-            val = replace_nan(corr_func(single_stat[:, 0], single_stat[0:, 1])[0], 0)
+            val = replace_nan(corr_func(single_stat[:, 0], single_stat[0:, 1])[0], 0.0)
         elif config.group_by == "sample":
             val = np.mean(single_stat)
         elif config.group_by == "system":
             n_systems = int(single_stat.shape[-1] / 2)
             true_scores = np.sum(single_stat[:, 0:n_systems], axis=0)
             pred_scores = np.sum(single_stat[:, n_systems:], axis=0)
-            val = replace_nan(corr_func(true_scores, pred_scores)[0], 0)
+            val = replace_nan(corr_func(true_scores, pred_scores)[0], 0.0)
         else:
             raise ValueError(
                 f"group_by with the value {config.group_by} hasn't been supported."
@@ -387,6 +386,5 @@ class PearsonCorrelationWMTDA(CorrelationWMTDAMetric):
 
         assert len(system_score) == len(manual_score)
 
-        res = stats.pearsonr(system_score, manual_score)[0]
-        val = 0 if math.isnan(res) else res
+        val = replace_nan(stats.pearsonr(system_score, manual_score)[0], 0.0)
         return val
--- a/explainaboard/metrics/meta_evaluation.py
+++ b/explainaboard/metrics/meta_evaluation.py
@@ -3,10 +3,10 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+import math
 from typing import Any, Optional, Union
 
 import numpy as np
-import math
 from scipy import stats
 
 from explainaboard.metrics.metric import (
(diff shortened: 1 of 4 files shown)


explainaboard/metrics/meta_evaluation.py
```diff
@@ -77,7 +78,9 @@ def calc_stats_from_data(
             return SimpleMetricStats(
                 np.array(
                     [
-                        corr_func(true, pred)[0]
+                        0
+                        if math.isnan(corr_func(true, pred)[0])
+                        else corr_func(true, pred)[0]
```
odashi, line 83: This calls the same function twice. this checking needs to be wrapped by a function, or corr functions need to return always valid number.

explainaboard/metrics/meta_evaluation.py
```diff
@@ -120,14 +123,16 @@ def _calc_metric_from_aggregate_single(self, single_stat: np.ndarray) -> float:
         config = narrow(CorrelationNLGConfig, self.config)
         corr_func = config.get_correlation_func(config.correlation_type)
         if config.group_by == "dataset":
-            val = corr_func(single_stat[:, 0], single_stat[0:, 1])[0]
+            res = corr_func(single_stat[:, 0], single_stat[0:, 1])[0]
+            val = 0 if math.isnan(res) else res
```
odashi, line 127: The same logic is implemented multiple times. Introducing a function is better as follows:
```python
def replace_nan(value: float, default: float) -> float:
    """docstring"""
    return default if math.isnan(value) else value
```

explainaboard/metrics/meta_evaluation.py
```diff
@@ -385,6 +387,6 @@ def calc_metric_from_aggregate_single(self, single_stat: np.ndarray) -> float:
 
         assert len(system_score) == len(manual_score)
 
-        val = stats.pearsonr(system_score, manual_score)[0]
-
+        res = stats.pearsonr(system_score, manual_score)[0]
+        val = 0 if math.isnan(res) else res
```
odashi, line 391: this would be replaced too

explainaboard/metrics/meta_evaluation.py
```diff
@@ -116,18 +118,18 @@ def _calc_metric_from_aggregate_single(self, single_stat: np.ndarray) -> float:
         Returns:
             The aggregated metric value.
         """
-        val = 0
+        val = 0.0
         config = narrow(CorrelationNLGConfig, self.config)
         corr_func = config.get_correlation_func(config.correlation_type)
         if config.group_by == "dataset":
-            val = corr_func(single_stat[:, 0], single_stat[0:, 1])[0]
+            val = replace_nan(corr_func(single_stat[:, 0], single_stat[0:, 1])[0], 0)
```
odashi, line 125:
```suggestion
            val = replace_nan(corr_func(single_stat[:, 0], single_stat[0:, 1])[0], 0.0)
```

explainaboard/metrics/meta_evaluation.py
```diff
@@ -116,18 +118,18 @@ def _calc_metric_from_aggregate_single(self, single_stat: np.ndarray) -> float:
...
         elif config.group_by == "sample":
             val = np.mean(single_stat)
         elif config.group_by == "system":
             n_systems = int(single_stat.shape[-1] / 2)
             true_scores = np.sum(single_stat[:, 0:n_systems], axis=0)
             pred_scores = np.sum(single_stat[:, n_systems:], axis=0)
-            val = corr_func(true_scores, pred_scores)[0]
+            val = replace_nan(corr_func(true_scores, pred_scores)[0], 0)
```
odashi, line 132:
```suggestion
            val = replace_nan(corr_func(true_scores, pred_scores)[0], 0.0)
```

explainaboard/utils/py_utils_test.py
```diff
@@ -0,0 +1,15 @@
...
+
+from __future__ import annotations
+
+import math
+import unittest
+
+from explainaboard.utils.py_utils import replace_nan
+
+
+class ReplaceNanTest(unittest.TestCase):
```
odashi, line 11:
```suggestion
class ReplaceNanTest(unittest.TestCase) -> None:
```

explainaboard/utils/py_utils_test.py
```diff
@@ -0,0 +1,15 @@
...
+    def test_replace_nan(self):
+        self.assertEqual(replace_nan(math.nan, 10), 10)
+
+        self.assertEqual(replace_nan(1, 10), 1)
```
odashi, line 15:
```suggestion
        self.assertEqual(replace_nan(math.nan, 10.0), 10.0)

        self.assertEqual(replace_nan(1.0, 10.0), 1.0)
```

explainaboard/utils/py_utils_test.py
pfliu-nlp: it's a little strange to put `-> None` at the end of the class, which will result in an error. So I put them at the end of the test function.

odashi: Sorry this is my bad. Please ignore this comment.

odashi (approved):

PR #605, opened 2022-12-27
add tests for meval to replicate paper results

pfliu-nlp:
# Overview

This PR adds tests to verify whether our implemented meta-evaluation processor is able to replicate reported results from existing published papers.

# Details

* Collect system outputs from this [repo](https://github.com/neulab/BARTScore) of two metrics (rouge1 and bartscore)
* Using Explainaboard to process these outputs and compare the results with the ones reported from the above repo.

# References

- Paper: BARTSCORE: Evaluating Generated Text as Text Generation
- Code: https://github.com/neulab/BARTScore


integration_tests/meta_eval_nlg_test.py
```diff
@@ -189,3 +182,74 @@ def test_dataset_level_spearmanr_bootstrap(self) -> None:
...
+        sys_info = processor.process(metadata, data)
+        overall_score = (
+            sys_info.results.overall["example"]["SpearmanSampleLevelCorr"]
+            .get_value(Score, "score")
+            .value
+        )
+        self.assertGreater(len(sys_info.results.analyses), 0)
+        self.assertAlmostEqual(
+            overall_score,
+            0.0946,
```
odashi, line 224: I couldn't find this score in the ROUGE results in the original paper.

integration_tests/meta_eval_nlg_test.py
```diff
@@ -189,3 +182,74 @@ def test_dataset_level_spearmanr_bootstrap(self) -> None:
...
+        sys_info = processor.process(metadata, data)
+        overall_score = (
+            sys_info.results.overall["example"]["SpearmanSampleLevelCorr"]
+            .get_value(Score, "score")
+            .value
+        )
+        self.assertGreater(len(sys_info.results.analyses), 0)
+        self.assertAlmostEqual(
+            overall_score,
+            0.3157,
```
odashi, line 253: Ditto

integration_tests/meta_eval_nlg_test.py
```diff
@@ -189,3 +182,74 @@ def test_dataset_level_spearmanr_bootstrap(self) -> None:
         ci = unwrap(corr_metric.calc_confidence_interval(stats, 0.05))
         self.assertAlmostEqual(ci[0], 1, 2)
         self.assertAlmostEqual(ci[1], 1, 2)
+
+
+class MetaEvalNLGNewsroomTest(unittest.TestCase):
+    """
+    Test the NLG metric on newsroom dataset and replicate the reported results from
+    the paper: https://arxiv.org/pdf/2106.11520.pdf
```
neubig, line 190: Could you please add some of the details from the PR into this comment here so they remain for later?

Also, I agree with Yusuke's comments below that it'd be a good idea to specify which results were replicated.

neubig (commented):
Thanks for this! Overall, I think this is a great initiative.

However, if the numbers are a little bit different because of the stemmer used for ROUGE, actually they aren't really "replicated" I guess? To truly replicate the results we can probably calculate ROUGE with the same stemmer as the original paper. Would that be hard?

integration_tests/meta_eval_nlg_test.py
pfliu-nlp: ROUGE1: Table 4 -> Column:  NeR18 -> COH

integration_tests/meta_eval_nlg_test.py
pfliu-nlp: This result didn't be listed in paper, but are stored in their github repo: implementation section: https://github.com/neulab/BARTScore#reproduce

integration_tests/meta_eval_nlg_test.py
pfliu-nlp: >"To truly replicate the results we can probably calculate ROUGE with the same stemmer as the original paper. Would that be hard?"

I won't plan to do this.  (1) We need to modify the eaas code. (2) the original paper use a package that will be outdated. no need to support it.  Anyway, no plan to do this here.

integration_tests/meta_eval_nlg_test.py
```diff
@@ -189,3 +182,77 @@ def test_dataset_level_spearmanr_bootstrap(self) -> None:
...
+        }
+        processor = get_processor_class(TaskType.meta_evaluation_nlg)()
+        sys_info = processor.process(metadata, data)
+        overall_score = (
+            sys_info.results.overall["example"]["SpearmanSampleLevelCorr"]
+            .get_value(Score, "score")
+            .value
+        )
+        self.assertGreater(len(sys_info.results.analyses), 0)
+        # Replicate the Table 4 result in paper: https://arxiv.org/pdf/2106.11520.pdf
```
neubig, line 222:
```suggestion
        # Replicate the Table 4 result in paper: https://arxiv.org/pdf/2106.11520.pdf
        # Specifically: ROUGE1: Table 4 -> Column: NeR18 -> COH
```

integration_tests/meta_eval_nlg_test.py
neubig: Thanks! I still didn't see where the number 0.3157 came from though. The Spearman numbers in the table were in the range 0.49.

Could you add a little bit more detail to the comment about where the number could be found?

Sorry, I don't want to be pedantic, but it'd be nice to keep this information around and if we don't do it know we'll likely forget later.

neubig (commented):
Thanks a lot! I was able to confirm the ROUGE number but still wasn't able to find the BARTScore number.

Also, I reviewed this before you clicked the "re-request review button", so if this was still WIP apologies!

PR #609, opened 2023-02-13
Add negative error

neubig:
# Overview

This adds the ability to flip error-based evaluation metrics such as MAE and RMSE to be negative, so that higher values are better.

It also adds tests for MAE and RMSE, which did not exist before.

# Blocked by

- NA


neubig:
(Integration tests are failing but they seem to be due to an unrelated issue with eaas.)

odashi (approved):

